Issue #3, opened 2019-02-26
Log equivocation reports


weilbith:
Another logging callback should be defined and registered to the equivocation reporter. This function should log all information requested in trustlines-network/project#299. Furthermore it should provide all data that are needed to call an equivocation proof for reporting the related validator, as defined in trustlines-network/project#300.

berndbohmeier:
Done in #4

PR #4, opened 2019-02-28
Add equivocation report logging functionality

weilbith:
**Closes:**
 - #3

**Related to:**
 - trustlines-network/project#299
 - trustlines-network/project#299

The logging has been tested by a local chain setup of two nodes, including one validator. The chain specification includes 5 validators, while 4 are offline, so no block reaches finality. The monitoring service connects to the non-validating node.
Both nodes are started, where the monitoring service is fetching the blocks from the non-validating node. At block `N` all nodes get stopped. The validators database gets deleted and the node restarted, sealing new blocks from genesis. As soon as the validator reaches `N+1` blocks, the non-validators connects again. The non-validator has to reorganize his chain, so the new blocks get fetched by the monitoring service as well. Equivocation reporter detects blocks by the same proposer for the same height (`N` times).

This is how such report looks like:
```
Proposer: 0xf59ca8dfed7f0c31feb86111a62a28260a1f1470
Block height: 1
Detection time: 2019-02-28 16:18:04.376685

Equivocated blocks:
0xa98ddece1763252c1e121e9b6fd60fbdb695c4de015a15abfb127596a4f5e0ba (2019-02-28 16:15:50)
0x68950e0a469c2285ada44cf3313ca7d3cbf203938c69db69180f0eaffeead841 (2019-02-28 16:17:05)

Data for an equivocation proof by the first two equivocated blocks:

RLP encoded block header one:
b'\xf9\x01\xf9\xa0\xab%\x90!\xa2\xd2\xc2\x07\x8a\xdaA\x153\xea\xca+\xfaq\x1f\xa6\xfe\xf3\x8ag\xcf\xcd\tc\x13F\xae*\xa0\x1d\xccM\xe8\xde\xc7]z\xab\x85\xb5g\xb6\xcc\xd4\x1a\xd3\x12E\x1b\x94\x8at\x13\xf0\xa1B\xfd@\xd4\x93G\x94\xf5\x9c\xa8\xdf\xed\x7f\x0c1\xfe\xb8a\x11\xa6*(&\n\x1f\x14p\xa02\xc3\xf6uIA\x14\x95\xcdG\xd1\x82\xdcC4\x89>5\xacF\x08\x08`Z\xb1\x87\x8f\x071\x8d\x10*\xa0V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!\xa0V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!\xb9\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x90\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xed\x81\x97\xc1\x01\x83z\x12\x00\x80\x84\\x\t6\x9f\xde\x83\x02\x03\x00\x8fParity-Ethereum\x861.30.0\x82li'

Signature of block header one:
0xc5102a8d8d3a9b03e000a9893df66e2ce933534694c46d3a3244a8931cfe8a2345a35f2c2e45e4aa54e1994a63b73dcca5703de625897b0fab45d2437755e4a100

RLP encoded block header two:
b'\xf9\x01\xf9\xa0\xab%\x90!\xa2\xd2\xc2\x07\x8a\xdaA\x153\xea\xca+\xfaq\x1f\xa6\xfe\xf3\x8ag\xcf\xcd\tc\x13F\xae*\xa0\x1d\xccM\xe8\xde\xc7]z\xab\x85\xb5g\xb6\xcc\xd4\x1a\xd3\x12E\x1b\x94\x8at\x13\xf0\xa1B\xfd@\xd4\x93G\x94\xf5\x9c\xa8\xdf\xed\x7f\x0c1\xfe\xb8a\x11\xa6*(&\n\x1f\x14p\xa02\xc3\xf6uIA\x14\x95\xcdG\xd1\x82\xdcC4\x89>5\xacF\x08\x08`Z\xb1\x87\x8f\x071\x8d\x10*\xa0V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!\xa0V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!\xb9\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x90\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xff\xed\x81\x97\xb2\x01\x83z\x12\x00\x80\x84\\x\t\x81\x9f\xde\x83\x02\x03\x00\x8fParity-Ethereum\x861.30.0\x82li'

Signature of block header two:
0xd906d6eb396903c02e2c972ffbb567a6a1461d6b866288bad9c03e247e86cf925806d8d224e825d674f32529fba13f22f4100a3a290d45066b2343430766766600
```

### DIFF
--- a/monitor/main.py
+++ b/monitor/main.py
@@ -44,6 +44,32 @@ MAX_REORG_DEPTH = (
     1000
 )  # blocks at this depth in the chain are assumed to not be replaced
 
+EQUIVOCATION_REPORT_TEMPLATE = """\
+Proposer: {proposer_address}
+Block height: {block_height}
+Detection time: {detection_time}
+
+Equivocated blocks:
+{block_hash_timestamp_list}
+
+Data for an equivocation proof by the first two equivocated blocks:
+
+RLP encoded block header one:
+{rlp_encoded_block_header_one}
+
+Signature of block header one:
+{signature_block_header_one}
+
+RLP encoded block header two:
+{rlp_encoded_block_header_two}
+
+Signature of block header two:
+{signature_block_header_two}
+
+------------------------------
+
+"""
+
 
 def step_number_to_timestamp(step):
     return step * STEP_DURATION
@@ -73,8 +99,6 @@ class App:
 
         self.state_path = self.db_dir / STATE_FILE_NAME
         self.skip_file = open(report_dir / SKIP_FILE_NAME, "a")
-
-        self.equivocation_report_counter = 1
 
         self.w3 = None
         self.get_primary_for_step = None
@@ -244,41 +268,39 @@ class App:
         block_one = get_canonicalized_block(self.w3.eth.getBlock(block_hash_one))
         block_two = get_canonicalized_block(self.w3.eth.getBlock(block_hash_two))
 
-        # All blocks share the same proposer and height, so just pick the first.
-        proposer = encode_hex(get_proposer(block_one))
-        height = block_two.number
+        block_hash_timestamp_list = ""
+
+        for block_hash in equivocated_block_hashes:
+            block = self.w3.eth.getBlock(block_hash)
+            block_hash_timestamp_list += "{} ({})\n".format(
+                encode_hex(block_hash),
+                datetime.datetime.utcfromtimestamp(block.timestamp),
+            )
+
+        proposer_address = encode_hex(get_proposer(block_one))
+
+        equivocation_report_template_variables = {
+            "proposer_address": proposer_address,
+            "block_height": block_one.number,
+            "detection_time": datetime.datetime.utcnow(),
+            "block_hash_timestamp_list": block_hash_timestamp_list,
+            "rlp_encoded_block_header_one": rlp_encoded_block(block_one),
+            "signature_block_header_one": keys.Signature(block_one.signature),
+            "rlp_encoded_block_header_two": rlp_encoded_block(block_two),
+            "signature_block_header_two": keys.Signature(block_two.signature),
+        }
+
+        equivocation_report_file_name = (
+            f"equivocation_reports_for_proposer_{proposer_address}"
+        )
 
         with open(
-            self.report_dir / f"equivocation_report_{self.equivocation_report_counter}",
-            "w",
-        ) as file:
-            self.equivocation_report_counter += 1
-
-            file.write(f"Proposer: {proposer}\n")
-            file.write(f"Height: {height}\n")
-            file.write(f"Detection time: {datetime.datetime.utcnow()}\n")
-
-            file.write("\nEquivocated blocks:\n")
-            for block_hash in equivocated_block_hashes:
-                block = self.w3.eth.getBlock(block_hash)
-                file.write(
-                    f"{encode_hex(block_hash)} ({datetime.datetime.utcfromtimestamp(block.timestamp)})\n"
+            self.report_dir / equivocation_report_file_name, "a"
+        ) as equivocation_report_file:
+            equivocation_report_file.write(
+                EQUIVOCATION_REPORT_TEMPLATE.format(
+                    **equivocation_report_template_variables
                 )
-
-            file.write(
-                "\n\nData for an equivocation proof by the first two equivocated blocks:\n"
-            )
-            file.write(
-                f"\nRLP encoded block header one:\n{rlp_encoded_block(block_one)}\n"
-            )
-            file.write(
-                f"\nSignature of block one:\n{keys.Signature(block_one.signature)}\n"
-            )
-            file.write(
-                f"\nRLP encoded block header two:\n{rlp_encoded_block(block_one)}\n"
-            )
-            file.write(
-                f"\nSignature of block two:\n{keys.Signature(block_two.signature)}\n"
             )
 
 
--- a/monitor/main.py
+++ b/monitor/main.py
@@ -252,15 +252,15 @@ class App:
     def equivocation_logger(self, equivocated_block_hashes):
         """Log a reported equivocation event.
 
-        Each new equivocation report is logged into a new file, using a counter
-        as unique identifier.
-        Logged information are the proposer of the blocks, the height at which
-        all blocks have been equivocated and a list of all block hashes with
-        their timestamp.
-        Additionally two representing blocks are logged with their RLP encoded
-        header and related signature, which can be used for an equivocation
-        proof on reporting a validator.
+        Equivocation reports are logged into files separated by the proposers
+        address. Logged information are the proposer of the blocks, the height
+        at which all blocks have been equivocated and a list of all block hashes
+        with their timestamp. Additionally two representing blocks are logged
+        with their RLP encoded header and related signature, which can be used
+        for an equivocation proof on reporting a validator.
         """
+
+        assert len(equivocated_block_hashes) >= 2
 
         blocks = [
             self.w3.eth.getBlock(block_hash) for block_hash in equivocated_block_hashes
--- a/monitor/main.py
+++ b/monitor/main.py
@@ -44,13 +44,14 @@ MAX_REORG_DEPTH = (
     1000
 )  # blocks at this depth in the chain are assumed to not be replaced
 
+BLOCK_HASH_AND_TIMESTAMP_TEMPLATE = "{block_hash} ({block_timestamp})"
 EQUIVOCATION_REPORT_TEMPLATE = """\
 Proposer: {proposer_address}
 Block height: {block_height}
 Detection time: {detection_time}
 
 Equivocated blocks:
-{block_hash_timestamp_list}
+{block_hash_timestamp_summary}
 
 Data for an equivocation proof by the first two equivocated blocks:
 
@@ -261,29 +262,31 @@ class App:
         proof on reporting a validator.
         """
 
-        # Use the first two blocks as representational data for the equivocation proof.
-        block_hash_one = equivocated_block_hashes[0]
-        block_hash_two = equivocated_block_hashes[1]
+        blocks = [
+            self.w3.eth.getBlock(block_hash) for block_hash in equivocated_block_hashes
+        ]
 
-        block_one = get_canonicalized_block(self.w3.eth.getBlock(block_hash_one))
-        block_two = get_canonicalized_block(self.w3.eth.getBlock(block_hash_two))
-
-        block_hash_timestamp_list = ""
-
-        for block_hash in equivocated_block_hashes:
-            block = self.w3.eth.getBlock(block_hash)
-            block_hash_timestamp_list += "{} ({})\n".format(
-                encode_hex(block_hash),
-                datetime.datetime.utcfromtimestamp(block.timestamp),
+        block_hashes_and_timestamp_strings = [
+            BLOCK_HASH_AND_TIMESTAMP_TEMPLATE.format(
+                block_hash=encode_hex(block.hash),
+                block_timestamp=datetime.datetime.utcfromtimestamp(block.timestamp),
             )
+            for block in blocks
+        ]
 
-        proposer_address = encode_hex(get_proposer(block_one))
+        block_hash_and_timestamp_summary = "\n".join(block_hashes_and_timestamp_strings)
+
+        # Use the first two blocks as representational data for the equivocation proof.
+        block_one = get_canonicalized_block(blocks[0])
+        block_two = get_canonicalized_block(blocks[1])
+
+        proposer_address_hex = encode_hex(get_proposer(block_one))
 
         equivocation_report_template_variables = {
-            "proposer_address": proposer_address,
+            "proposer_address": proposer_address_hex,
             "block_height": block_one.number,
             "detection_time": datetime.datetime.utcnow(),
-            "block_hash_timestamp_list": block_hash_timestamp_list,
+            "block_hash_timestamp_summary": block_hash_and_timestamp_summary,
             "rlp_encoded_block_header_one": rlp_encoded_block(block_one),
             "signature_block_header_one": keys.Signature(block_one.signature),
             "rlp_encoded_block_header_two": rlp_encoded_block(block_two),
@@ -291,7 +294,7 @@ class App:
         }
 
         equivocation_report_file_name = (
-            f"equivocation_reports_for_proposer_{proposer_address}"
+            f"equivocation_reports_for_proposer_{proposer_address_hex}"
         )
 
         with open(
--- a/monitor/main.py
+++ b/monitor/main.py
@@ -13,13 +13,14 @@ from sqlalchemy import create_engine
 
 from web3 import Web3, HTTPProvider
 from eth_utils import encode_hex
+from eth_keys import keys
 
 from monitor.db import BlockDB
 from monitor.block_fetcher import BlockFetcher
 from monitor.offline_reporter import OfflineReporter
 from monitor.skip_reporter import SkipReporter
 from monitor.equivocation_reporter import EquivocationReporter
-
+from monitor.blocks import get_canonicalized_block, get_proposer, rlp_encoded_block
 from monitor.validators import make_primary_function
 
 import click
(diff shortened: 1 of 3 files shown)


monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+        """
+
+        # Use the first two blocks as representational data for the equivocation proof.
+        block_hash_one = equivocated_block_hashes[0]
+        block_hash_two = equivocated_block_hashes[1]
+
+        block_one = get_canonicalized_block(self.w3.eth.getBlock(block_hash_one))
+        block_two = get_canonicalized_block(self.w3.eth.getBlock(block_hash_two))
+
+        # All blocks share the same proposer and height, so just pick the first.
```
jannikluhn: `assert`s are stronger than comments:
```suggestion
        assert block_one.proposer == block_two.proposer and block_one.height == block_two.height
```

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+        proposer = encode_hex(get_proposer(block_one))
```
jannikluhn:
```suggestion
        proposer = get_proposer(block_one)
```

as from the variable name I wouldn't expect `proposer` to be hex encoded

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+        height = block_two.number
+
+        with open(
+            self.report_dir / f"equivocation_report_{self.equivocation_report_counter}",
+            "w",
+        ) as file:
```
jannikluhn: `file` is a built in function, so you should rename it (maybe `equivocation_file` or something)

monitor/main.py
```diff
@@ -73,6 +74,8 @@ def __init__(
         self.state_path = self.db_dir / STATE_FILE_NAME
         self.skip_file = open(report_dir / SKIP_FILE_NAME, "a")
 
+        self.equivocation_report_counter = 1
```
jannikluhn: This will get reset on restarts. So either put it in the application state or, what I'd probably prefer, get rid of it completely and just create filenames until you find one that doesn't exist. Or create one file per validator and append  the equivocations.

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+
+        block_one = get_canonicalized_block(self.w3.eth.getBlock(block_hash_one))
+        block_two = get_canonicalized_block(self.w3.eth.getBlock(block_hash_two))
+
+        # All blocks share the same proposer and height, so just pick the first.
+        proposer = encode_hex(get_proposer(block_one))
+        height = block_two.number
+
+        with open(
+            self.report_dir / f"equivocation_report_{self.equivocation_report_counter}",
```
jannikluhn: Maybe put the validator address into the filename.

monitor/main.py
jannikluhn: If you decide to keep it, starting with 0 would make more sense in my opinion.

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+            "w",
+        ) as file:
+            self.equivocation_report_counter += 1
```
jannikluhn: Maybe increment this at the beginning of the method.

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+
+            file.write(f"Proposer: {proposer}\n")
```
jannikluhn: I think constructing the lines first and only after that then writing them to the file would be more readable and shorten the `with` block. E.g.:

```Python
lines = [
    f"Height: {height}",
    ...
]

with open() as asdf:
    asdf.writelines(lines)
```

You could also consider defining the format in a big template block at the beginning and only filling it here (using `EQUIVOCATION_TEMPLATE.format(height=height, ...)`), but I'm not sure how well this would work with the block hash loop.

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+
+            file.write("\nEquivocated blocks:\n")
+            for block_hash in equivocated_block_hashes:
+                block = self.w3.eth.getBlock(block_hash)
+                file.write(
+                    f"{encode_hex(block_hash)} ({datetime.datetime.utcfromtimestamp(block.timestamp)})\n"
+                )
+
+            file.write(
+                "\n\nData for an equivocation proof by the first two equivocated blocks:\n"
```
jannikluhn: Maybe separate the line breaks at the beginning into it's own `file.write` (assuming the `file.write` pattern stays)

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+            )
+            file.write(
+                f"\nRLP encoded block header one:\n{rlp_encoded_block(block_one)}\n"
+            )
+            file.write(
+                f"\nSignature of block one:\n{keys.Signature(block_one.signature)}\n"
```
jannikluhn:
```suggestion
                f"\nSignature of block header one:\n{keys.Signature(block_one.signature)}\n"
```

monitor/main.py
```diff
@@ -220,6 +224,63 @@ def offline_logger(self, validator, steps):
...
+            )
+            file.write(
+                f"\nRLP encoded block header two:\n{rlp_encoded_block(block_one)}\n"
+            )
+            file.write(
+                f"\nSignature of block two:\n{keys.Signature(block_two.signature)}\n"
```
jannikluhn:
```suggestion
                f"\nSignature of block header two:\n{keys.Signature(block_two.signature)}\n"
```

monitor/main.py
weilbith: Okay, I thought about it. The point is, that actually the meaning is that **all** blocks have the same proposer and height. To `assert` this, I would need to query all blocks and assert them. Do you think using the first two representing blocks is enough?

monitor/main.py
weilbith: Sry, I wanted to make it compliant with your last review request.
Is it rather about a change of the variable name or do we really want to log the address in bytes, which is  not well readable?

monitor/main.py
weilbith: Oh, that is true!!

monitor/main.py
weilbith: I thought we have agreed on this design at the meeting. Your concerns are valid! Since I still think having a file per report is more nice, I would go for you first suggested solution on finding the next free filename. Not sure how efficient this is, since this would increase the looping by one round for each new report.
Maybe I should try to get all files ordered (which would be done by the number), take the last entry, parse its number and increment it.
Does this sounds like a well suited solution? E

monitor/main.py
weilbith: Since we replace this functionality by the above discussed requests, I guess this here will be obsolete.

monitor/main.py
weilbith: Nice idea. Also thought this must be maybe look more nice. Wanted do somehow display the file structure. You're right, the loop is the "problem" here. But it should be possible. Let me try do write something.

monitor/main.py
weilbith: > assuming the `file.write` pattern stays

Yes, will not be the case, by the above requests. Can be resolved as soon as I implemented the alternative design.

monitor/main.py
weilbith: When I just read this, I possibly favor the single file per validator approach. Even if it is not thaaat nice, I think constructing the above solution makes the code too complicated for this simple problem. DO you agree? :thinking:

monitor/main.py
weilbith: Just as reference: For the offline logger you are also using the hex encoded variation.

monitor/main.py
weilbith: Change part of the most recent version.

monitor/main.py
weilbith: I implemented the last suggested approach now, due my last comments. Its functionality has been tested. Hope this design is fine.

monitor/main.py
weilbith: Is the case now, since having one file per proposer.

monitor/main.py
weilbith: If you like to check my last commit, I introduced a template as you suggested and flush all information with a single write instruction. Please check if you like this more. :upside_down_face:

monitor/main.py
weilbith: Is part of the new template.

monitor/main.py
weilbith: Is part of the new template.

monitor/main.py
```diff
@@ -220,6 +248,61 @@ def offline_logger(self, validator, steps):
...
+        """
+
+        # Use the first two blocks as representational data for the equivocation proof.
+        block_hash_one = equivocated_block_hashes[0]
+        block_hash_two = equivocated_block_hashes[1]
+
+        block_one = get_canonicalized_block(self.w3.eth.getBlock(block_hash_one))
+        block_two = get_canonicalized_block(self.w3.eth.getBlock(block_hash_two))
+
+        block_hash_timestamp_list = ""
```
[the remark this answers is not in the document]
weilbith: For this looks more bloated and there less readable, but I like to take the suggestion. Not exactly sure, why using a global template only for the hash and timestamp as well. I mean we don't do this for every little formatting, don't we? :thinking: 
Yeah, naming is hard. Was also not happy, but at least they say what they are meant for.

> Just two additional suggestions, but other than that it's fine

It would be awesome, if you could resolve the old/other requests, if you think they are done. Thanks!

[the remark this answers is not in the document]
jannikluhn: Don't do it if you don't like it, but I think it's better as it first does one task and then the next one, instead of interleaving the two (and it doesn't mutate any variables).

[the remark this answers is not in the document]
weilbith: That is true. The code looks nice. Even if it are more lines of code, I think it is the better one. Also I recognized I can reuse this to get the representational blocks. So this isn't that much more. So I think the flow is clear: fetch all blocks, extract the information, build the hash-list-thing, construct the variable object by all information and write to a file.
For it is readable this way. At least it is a whole bunch of information logged compared to the other logger functions.
Update it very soon. Just need to test it again.

monitor/main.py
```diff
@@ -220,6 +248,61 @@ def offline_logger(self, validator, steps):
...
+
+        for block_hash in equivocated_block_hashes:
+            block = self.w3.eth.getBlock(block_hash)
+            block_hash_timestamp_list += "{} ({})\n".format(
+                encode_hex(block_hash),
+                datetime.datetime.utcfromtimestamp(block.timestamp),
+            )
+
+        proposer_address = encode_hex(get_proposer(block_one))
```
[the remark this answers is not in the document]
weilbith: Has been done during the updates for the above request. :+1:

monitor/main.py
```diff
@@ -220,6 +249,63 @@ def offline_logger(self, validator, steps):
...
+                block_timestamp=datetime.datetime.utcfromtimestamp(block.timestamp),
+            )
+            for block in blocks
+        ]
+
+        block_hash_and_timestamp_summary = "\n".join(block_hashes_and_timestamp_strings)
+
+        # Use the first two blocks as representational data for the equivocation proof.
+        block_one = get_canonicalized_block(blocks[0])
+        block_two = get_canonicalized_block(blocks[1])
```
berndbohmeier: Maybe add a 
`assert len(equivocated_block_hashes) >=2 `
at the beginning of this function to fail early if used wrong

monitor/main.py
```diff
@@ -220,6 +249,63 @@ def offline_logger(self, validator, steps):
         with open(self.report_dir / filename, "w") as f:
             json.dump({"validator": encode_hex(validator), "missed_steps": steps}, f)
 
+    def equivocation_logger(self, equivocated_block_hashes):
+        """Log a reported equivocation event.
+
+        Each new equivocation report is logged into a new file, using a counter
```
berndbohmeier: Seems to be outdated information, please update.

weilbith: Well observed! :+1:

monitor/main.py
weilbith: Should I make a comment for the purpose or is it self-explaining?

berndbohmeier: self explaining i think

PR #7, opened 2019-03-11
do not use asyncio anymore

schmir:
The use of asyncio didn't really serve any purpose. In fact it was preventing
the signal handler to be run immediately.

fixes https://github.com/trustlines-network/tlbc-monitor/issues/6

### DIFF
--- a/monitor/main.py
+++ b/monitor/main.py
@@ -1,9 +1,9 @@
-import asyncio
 import datetime
 import json
 from pathlib import Path
 import pickle
 import signal
+import time
 
 from typing import Any, NamedTuple
 
@@ -117,7 +117,7 @@ class App:
         self._register_reporter_callbacks()
         self._running = False
 
-    async def run(self):
+    def run(self):
         self._running = True
         try:
             self.logger.info("starting sync")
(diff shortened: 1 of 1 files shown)


schmir:
> But on initial synchronization this could take since the window is quite large at this point. Correct?

yes, that's correct. And it's the reason I do think we should make this window much smaller.

monitor/main.py
```diff
@@ -131,15 +131,15 @@ def __init__(
...
-                    await asyncio.sleep(0.01)
+                    time.sleep(BLOCK_FETCH_INTERVAL)
         finally:
             self.skip_file.close()
             self.dump_app_state()
 
     def stop(self):
-        self.logger.info("Stopping... ")
+        self.logger.info(
+            "Stopping tlbc-monitor. This may take a long time, please be patient!"
```
jannikluhn: Just a note: We do have some parameter (`max_number_of_blocks`) that we can adjust to get an acceptable time again. But I'm fine with leaving it like this.

schmir: Thanks for the note. I guess we will change that eventually.

PR #9, opened 2019-03-12
Fix/mypy

schmir:
This runs mypy on circleci and works around a mypy error message. This is based on PR #8


schmir:
We already had some type annotations before, that's why I've added it to the circleci config.

monitor/db.py
```diff
@@ -10,8 +11,7 @@
 
 from monitor.blocks import get_canonicalized_block, get_proposer
 
-
-Base = declarative_base()
+Base: Any = declarative_base()
```
jannikluhn: Is there a reason for adding this particular type hint and this one only?

schmir:
sorry, for requesting a second review. nothing changed, I just wanted to try out that button.

monitor/db.py
jannikluhn: Is there a reason for adding this particular type hint and this one only?

schmir: see commit 972b7365ee59ed12d53cda092f3d7e

schmir:
jannikluhn <notifications@github.com> writes:

> jannikluhn commented on this pull request.
>
>
>
>> @@ -10,8 +11,7 @@
>
>  from monitor.blocks import get_canonicalized_block, get_proposer
>
> -
> -Base = declarative_base()
> +Base: Any = declarative_base()
>
> Is there a reason for adding this particular type hint and this one only?


yes,  the commit message says:

    workaround mypy 'invalid base class' issue
    
    Otherwise mypy complains 'error: Invalid base class' for this code:
    
    ,----
    | class Block(Base):
    |     __tablename__ = "blocks"
    | ...
    `----
    
    see https://github.com/python/mypy/issues/4284

PR #13, opened 2019-03-13
Fix/store state in db

schmir:
store pickled state in the database

This fixes https://github.com/trustlines-network/tlbc-monitor/issues/5 and https://github.com/trustlines-network/tlbc-monitor/issues/1

### DIFF
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -62,8 +62,6 @@ class BlockFetcher:
 
         try:
             self.db.insert_branch(blocks)
-            # head and current branch will be saved as part of the app state,
-            # change them before storing the app state
             self.head = blocks[-1]
             self.current_branch.clear()
         except AlreadyExists:
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -41,11 +41,23 @@ class BlockFetcher:
     def register_report_callback(self, callback):
         self.report_callbacks.append(callback)
 
+    def _run_callbacks(self, blocks):
+        for block in blocks:
+            # XXX block_fetcher currently does no callbacks for the first block
+            # see https://github.com/trustlines-network/tlbc-monitor/issues/12
+            # we reproduce that bug here on purpose for the moment (i.e. too
+            # lazy to fix the tests)
+            if block.number == 0:
+                continue
+
+            for callback in self.report_callbacks:
+                callback(block)
+
     def _insert_branch(self, blocks):
         if len(blocks) == 0:
             return
 
-        if not self.db.contains(blocks[0].parentHash):
+        if blocks[0].number != 0 and not self.db.contains(blocks[0].parentHash):
             raise ValueError("Tried to insert block with unknown parent")
 
         try:
@@ -53,14 +65,14 @@ class BlockFetcher:
         except AlreadyExists:
             raise ValueError("Tried to insert already known block")
 
-        for block in blocks:
-            for callback in self.report_callbacks:
-                callback(block)
+        self._run_callbacks(blocks)
+
+        self.head = blocks[-1]
+        self.current_branch.clear()
 
     def _insert_first_block(self):
         block = self.w3.eth.getBlock(0)
-        self.head = block
-        self.db.insert(block)
+        self._insert_branch([block])
 
     def fetch_and_insert_new_blocks(self, max_number_of_blocks=5000):
         """Fetches up to `max_number_of_blocks` blocks and updates the internal state
@@ -114,10 +126,7 @@ class BlockFetcher:
             )
         )
 
-        if len(blocks) > 0:
-            self._insert_branch(blocks)
-            self.head = blocks[-1]
-
+        self._insert_branch(blocks)
         return len(blocks)
 
     def _sync_backwards(self, max_blocks_to_fetch):
@@ -127,8 +136,6 @@ class BlockFetcher:
 
         if complete and len(self.current_branch) > 0:
             self._insert_branch(list(reversed(self.current_branch)))
-            self.head = self.current_branch[0]
-            self.current_branch.clear()
 
         return number_of_fetched_blocks
 
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -3,7 +3,7 @@ from typing import Any, NamedTuple
 
 import structlog
 
-from monitor.db import AlreadyExists
+from monitor.db import AlreadyExists, ensure_branch, blocks_from_block_dicts
 
 
 class BlockFetcherStateV1(NamedTuple):
@@ -16,7 +16,9 @@ class BlockFetcher:
 
     logger = structlog.get_logger("monitor.block_fetcher")
 
-    def __init__(self, state, w3, db, max_reorg_depth=1000):
+    def __init__(
+        self, state, w3, db, max_reorg_depth=1000, store_app_state=lambda _session: None
+    ):
         self.w3 = w3
         self.db = db
         self.max_reorg_depth = max_reorg_depth
@@ -25,6 +27,7 @@ class BlockFetcher:
         self.current_branch = state.current_branch
 
         self.report_callbacks = []
+        self.store_app_state = store_app_state
 
     @classmethod
     def from_fresh_state(cls, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


monitor/db.py
```diff
@@ -34,6 +35,54 @@ class Block(Base):
...
+def load_pickled(session, name):
+    """load the pickled NamedBlob object with the given name"""
+    named_blob = session.query(NamedBlob).filter_by(name=name).first()
+    if named_blob is None:
+        return None
+    else:
+        return pickle.loads(named_blob.blob)
+
+
+def store_pickled(session, name, obj):
```
berndbohmeier: We have now the old BlockDb as a interface for the db, and the new ralf style functions. This is not good, please go with one approach. I would like that the db code does not leak into the rest of the codebase, so please use proper abstraction.

monitor/block_fetcher.py
```diff
@@ -41,26 +44,47 @@ def state(self):
...
+        if blocks[0].number != 0 and not self.db.contains(blocks[0].parentHash):
             raise ValueError("Tried to insert block with unknown parent")
 
+        ensure_branch(blocks)
+        db_blocks = blocks_from_block_dicts(blocks)
+
+        session = self.db.session_class()
         try:
-            self.db.insert_branch(blocks)
+            session.add_all(db_blocks)
```
berndbohmeier: I dont want db code in here, that breaks the idea of the db abstraction and complicates things.

berndbohmeier: why not just 
```
insert_blocks(session, blocks)
```

monitor/block_fetcher.py
```diff
@@ -41,26 +44,47 @@ def state(self):
...
+            # head and current branch will be saved as part of the app state,
```
berndbohmeier: This is spaghetti code

monitor/db.py
berndbohmeier: If you want to allow that we can chain arbitrary db functions, you apparently dont want to use the db abstraction we tried to establish, so please refactor it, because it will not work together with your code.

monitor/block_fetcher.py
```diff
@@ -16,7 +16,9 @@ class BlockFetcher:
 
     logger = structlog.get_logger("monitor.block_fetcher")
 
-    def __init__(self, state, w3, db, max_reorg_depth=1000):
+    def __init__(
+        self, state, w3, db, max_reorg_depth=1000, store_app_state=lambda _session: None
```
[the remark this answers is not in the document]
schmir: that would be nice,but it's hard to do with the current code.

monitor/block_fetcher.py
schmir: Our db abstraction is broken, because it breaks atomicity. I've already written about that.

I'm fine with changing the methods to take an additional session argument or store a current session inside the object.
 
But this is a different issue. I couldn't use BlockDB for my purposes, therefore we've got some new functions here.

monitor/db.py
```diff
@@ -34,6 +35,54 @@ class Block(Base):
...
+    """store the given python obj as pickled NamedBlob object under the given name"""
+    pickled_state = pickle.dumps(obj)
+    named_blob = session.query(NamedBlob).filter_by(name=name).first()
```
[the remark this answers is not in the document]
schmir: thanks,will change that.

monitor/block_fetcher.py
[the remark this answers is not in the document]
schmir: the store_app_state thing would not be necessary, if the block_fetcher could update it's state on it's own in the database. But that was not possible with the current implementation. I've used what was there. I also don't like the use of pickle. But again, this is a different issue.

monitor/block_fetcher.py
berndbohmeier: > Our db abstraction is broken, because it breaks atomicity. I've already written about that.

That is not true, the automicity was broken because we had different places where we stored the state. The db by itself was fine. The current abstraction was not meant to do what we want to do now, so we should change it. 

> I'm fine with changing the methods to take an additional session argument or store a current session inside the object.

The internal state of the db is the session class. If you want to handle that outside we dont need the class anymore, so I would just remove it and do it like you do it with functions that take the session.

schmir: > That is not true, the automicity was broken because we had different places where we stored the state.

Sorry for not being clear enough. This is not what I meant.

What I did mean, is there is no way to do multiple database operations in a single transaction resulting in one atomic change to the database.
E.g. there is no way to insert two branches into the database in an atomic way with the current BlockDB implementation, because BlockDB manages the scope of the session (and transaction). 

The [sqlalchemy documentation on session management](https://docs.sqlalchemy.org/en/latest/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it) says you should not do that. In fact they have an example that looks like our code with '**Don't do this**' and 'this is the **wrong way to do it**' written above it.

monitor/block_fetcher.py
```diff
@@ -41,26 +41,39 @@ def state(self):
...
         if len(blocks) == 0:
             return
 
-        if not self.db.contains(blocks[0].parentHash):
+        if blocks[0].number != 0 and not self.db.contains(blocks[0].parentHash):
             raise ValueError("Tried to insert block with unknown parent")
 
         try:
             self.db.insert_branch(blocks)
+            # head and current branch will be saved as part of the app state,
```
berndbohmeier: Is this comment still valid? I dont see the storage of the app state here.

monitor/main.py
```diff
@@ -199,22 +196,7 @@ def _initialize_reporters(self, state_path, skip_rate, offline_window_size):
...
-                self.logger.critical("error loading state file")
-                raise
-
-        if not isinstance(app_state, AppStateV1):
-            raise RuntimeError("state file has unexpected format")
-
-        return app_state
-
     def _initialize_app_state(self):
-        self.logger.info("no state file found, starting from genesis")
```
berndbohmeier: Can we somewhere get this message back? I found it really helpful to know if it recovered from a state file or started from scratch.

monitor/block_fetcher.py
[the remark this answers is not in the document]
berndbohmeier: looks like this is done, please reopen if not.

monitor/block_fetcher.py
berndbohmeier: Yes, I would call that not flexible enough. We had a call and this is fixed now.

monitor/main.py
schmir: yes, I can change that. 

Actually I thought about having an explicit initialization step, which would create the database file.  I'd say the choice of using relative path names as a default together with automatically creating the  database isn't that good. I may open another issue for that.

monitor/block_fetcher.py
schmir: I will remove it.

tests/test_db.py
```diff
@@ -139,3 +139,50 @@ def test_empty_db_is_empty(empty_db):
...
+    assert empty_db.load_pickled("foo") == dict(bar=1)
+
+
+def test_store_pickled_existing(empty_db):
+    empty_db.store_pickled("foo", dict(bar=1))
+    empty_db.store_pickled("foo", dict(bar=2))
+    assert empty_db.load_pickled("foo") == dict(bar=2)
+
+
+def test_persisten_session(empty_db):
```
jannikluhn:
```suggestion
def test_persistent_session(empty_db):
```

monitor/db.py
```diff
@@ -34,50 +36,111 @@ class Block(Base):
...
+        ensure_branch(block_dicts)
+        blocks = blocks_from_block_dicts(block_dicts)
+        session = self._get_session()
         session.add_all(blocks)
 
         try:
-            session.commit()
+            session.flush()
+            if self.current_session is None:
+                session.commit()
```
jannikluhn: This confused me on my first read. Do we really need this or can we just require every change to be inside a `persistent_session` block?

monitor/db.py
```diff
@@ -34,50 +36,111 @@ class Block(Base):
...
         query = session.query(Block).filter(
             Block.proposer == proposer, Block.height == height
         )
         return query.all()
+
+    def store_pickled(self, name, obj):
+        session = self._get_session()
+        store_pickled(session, name, obj)
+        if self.current_session is None:
+            session.commit()
```
jannikluhn: see above

monitor/db.py
schmir: Well, the tests don't have explicit session handing, so this is mostly needed for the tests. We could maybe introduce a autouse fixture which opens a persistent connection for each test.

jannikluhn: I think this would be a little cleaner, but I don't feel too strongly about it. Your call :)

Issue #16, opened 2019-03-19
Report same slot instead of same height as equivocation 


berndbohmeier:
Belongs to: 

https://github.com/trustlines-network/project/issues/401
https://github.com/trustlines-network/project/issues/400

weilbith:
@jannikluhn is there anything left since #17 has been merged?

jannikluhn:
No, that should have been it.

weilbith:
Awkward that auto-closing hasn't worked out. :man_shrugging:

PR #17, opened 2019-03-20
Feature/step in db

### LINKED ISSUE #15: Include step in database schema
We have to include the step so that we can query for equivocated blocks. 
Change the methods that find equivocated blocks. 

Belongs to https://github.com/trustlines-network/project/issues/400
https://github.com/trustlines-network/project/issues/401

jannikluhn:
Changes implemented on the most recent @schmir branch I could find as I assume that they will be merged earlier.

Closes #15 and #16

Tests pass, but not yet run against actual parity node.


schmir:
> I've rebased that first commit from my branch and amended it with some other changes. Can we remove that here?

This can easily be revomed with git rebase. please remove it.

PR #21, opened 2019-03-29
Check for reorgs based on block before initial one

### LINKED ISSUE #20: Handle reorg on block before the selected initial block
The monitoring service does not handle reorgs on blocks which are based on a block that is older than the initial block the service has started to synchronize from (see `--sync-from` and its default).
As current behavior the service would start to sync backwards until reaching the genesis block. Afterwards (takes very long) it will fail without any meaningful error message. This must be caught earlier and with some useful output.

Related to:
- trustlines-network/project#356

weilbith:
Closes #20

Please help me with better names for the exception and improvement of the logging information. :expressionless:

### DIFF
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -97,11 +97,7 @@ class BlockFetcher:
 
     def _insert_first_block(self):
         resolver = self.initial_block_resolver or blocksel.ResolveGenesisBlock()
-        block = (
-            self.w3.eth.getBlock(self.initial_blocknr)
-            if self.initial_blocknr
-            else resolver.resolve_block(self.w3)
-        )
+        block = resolver.resolve_block(self.w3)
 
         if not block:
             raise ValueError("Can't fetch initial block to sync from!")
@@ -154,9 +150,6 @@ class BlockFetcher:
 
     def fetch_forward_sync_target(self):
         return max(self.w3.eth.blockNumber - self.max_reorg_depth, 0)
-
-    def _should_sync_forwards(self, current_block_number):
-        return current_block_number - self.head.number > self.max_reorg_depth
 
     def _sync_forwards(self, max_number_of_blocks):
         block_numbers_to_fetch = range(
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -11,6 +11,7 @@ from monitor import blocksel
 class BlockFetcherStateV1(NamedTuple):
     head: Any
     current_branch: Any
+    initial_blocknr: int
 
 
 class FetchedBlockBeforeInitialOneError(Exception):
@@ -41,7 +42,7 @@ class BlockFetcher:
 
         self.report_callbacks = []
         self.initial_block_resolver = initial_block_resolver
-        self.initial_blocknr = 0
+        self.initial_blocknr = state.initial_blocknr
 
     @classmethod
     def from_fresh_state(cls, *args, **kwargs):
@@ -49,11 +50,15 @@ class BlockFetcher:
 
     @classmethod
     def get_fresh_state(cls):
-        return BlockFetcherStateV1(head=None, current_branch=[])
+        return BlockFetcherStateV1(head=None, current_branch=[], initial_blocknr=0)
 
     @property
     def state(self):
-        return BlockFetcherStateV1(head=self.head, current_branch=self.current_branch)
+        return BlockFetcherStateV1(
+            head=self.head,
+            current_branch=self.current_branch,
+            initial_blocknr=self.initial_blocknr,
+        )
 
     def register_report_callback(self, callback):
         self.report_callbacks.append(callback)
@@ -90,7 +95,11 @@ class BlockFetcher:
 
     def _insert_first_block(self):
         resolver = self.initial_block_resolver or blocksel.ResolveGenesisBlock()
-        block = resolver.resolve_block(self.w3)
+        block = (
+            self.w3.eth.getBlock(self.initial_blocknr)
+            if self.initial_blocknr
+            else resolver.resolve_block(self.w3)
+        )
         latest = self.w3.eth.getBlock("latest")
         safe_initial_blocknr = max(latest.number - self.max_reorg_depth, 0)
         if block.number > safe_initial_blocknr:
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -14,7 +14,7 @@ class BlockFetcherStateV1(NamedTuple):
     initial_blocknr: int
 
 
-class FetchedBlockBeforeInitialOneError(Exception):
+class FetchingForkWithUnkownBaseError(Exception):
     pass
 
 
@@ -82,7 +82,9 @@ class BlockFetcher:
         if blocks[0].number not in (0, self.initial_blocknr) and not self.db.contains(
             blocks[0].parentHash
         ):
-            raise ValueError("Tried to insert block with unknown parent")
+            raise FetchingForkWithUnkownBaseError(
+                "Tried to insert branch from a fork with unknown parent block."
+            )
 
         try:
             self.db.insert_branch(blocks)
@@ -206,9 +208,11 @@ class BlockFetcher:
 
             if parent.number < self.initial_blocknr:
                 self.logger.error(
-                    f"Fetched block with number {parent.number} < {self.initial_blocknr} (initial block number)!"
+                    f"Fetched block with number {parent.number} < {self.initial_blocknr} (initial block number) on syncing backwards!"
                 )
-                raise FetchedBlockBeforeInitialOneError()
+                raise FetchingForkWithUnkownBaseError(
+                    "Synchronized backwards on a fork with base before initial synchronized block!"
+                )
 
             self.current_branch.append(parent)
 
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -100,6 +100,10 @@ class BlockFetcher:
             if self.initial_blocknr
             else resolver.resolve_block(self.w3)
         )
+
+        if not block:
+            raise ValueError("Can't fetch initial block to sync from!")
+
         latest = self.w3.eth.getBlock("latest")
         safe_initial_blocknr = max(latest.number - self.max_reorg_depth, 0)
         if block.number > safe_initial_blocknr:
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -13,6 +13,10 @@ class BlockFetcherStateV1(NamedTuple):
     current_branch: Any
 
 
+class FetchedBlockBeforeInitialOneError(Exception):
+    pass
+
+
 def format_block(block):
     dt = datetime.datetime.utcfromtimestamp(block.timestamp).replace(
         tzinfo=datetime.timezone.utc
@@ -186,6 +190,13 @@ class BlockFetcher:
         number_of_fetched_blocks = 0
         while not self.db.contains(self.current_branch[-1].parentHash):
             parent = self.w3.eth.getBlock(self.current_branch[-1].parentHash)
+
+            if parent.number < self.initial_blocknr:
+                self.logger.error(
+                    f"Fetched block with number {parent.number} < {self.initial_blocknr} (initial block number)!"
+                )
+                raise FetchedBlockBeforeInitialOneError()
+
             self.current_branch.append(parent)
 
             number_of_fetched_blocks += 1
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -179,12 +179,28 @@ class BlockFetcher:
 
         return number_of_fetched_blocks
 
+    def _get_block(self, blocknr_or_hash):
+        """call self.w3.eth.getBlock, but make sure we don't fetch a block
+        before the initial block"""
+        block = self.w3.eth.getBlock(blocknr_or_hash)
+        assert block is not None, f"Could not fetch block {blocknr_or_hash}"
+
+        if block.number < self.initial_blocknr:
+            self.logger.error(
+                f"Fetched block with number {block.number} < {self.initial_blocknr} (initial block number) on syncing backwards!"
+            )
+            raise FetchingForkWithUnkownBaseError(
+                "Synchronized backwards on a fork with base before initial synchronized block!"
+            )
+
+        return block
+
     def _fetch_branch(self, max_blocks_to_fetch):
         if max_blocks_to_fetch <= 0:
             raise ValueError("Maximum number of blocks to fetch must be positive")
 
         if len(self.current_branch) == 0:
-            head = self.w3.eth.getBlock("latest")
+            head = self._get_block("latest")
             if self.db.contains(head.hash):
                 self.logger.info(
                     "no new blocks",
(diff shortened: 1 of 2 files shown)


weilbith:
@schmir
Yes, this is such scenario. I guess he tries to find the parent for the genesis block here. So I take your suggestion and parametrize it.

I agree that the `run` method is kinda messy. But I thought handling the exception here makes sense, cause the `self._running` variable is here and allows a better overview of the workflow. To improve this, I'm want to make sure to understand you correctly. With `I'd rather not do it at all or...` you mean just throwing an exception that get not caught and terminate the service by this?

Oh, the restart problem is a very important point!! :expressionless:

schmir:
> With I'd rather not do it at all or... you mean just throwing an exception that get not caught and terminate the service by this?

yes, we do this for every other exception as well (i.e. if parity goes down). I'd rather have cleaner code in the run method and 'messy' exception reporting than the opposite. We can cleanup the exception handling later.

weilbith:
Do you think I should add another error message where the exception is thrown? Or should I replace it? Because I think the information that a block before the initial one is fetched does not tell users enough. :thinking:

weilbith:
> I guess he tries to find the parent for the genesis block here

My guess was wrong. It does not work, cause due to the `fetch_forward_sync_target` makes the service to sync forwards the second time the fetcher is called. This lead to the same problem of the fork, but is not handled. I'm a bit confused why it starts with backward syncing and then go for forward syncing afterwards. I mean this is related to the awkward test scenario of a pretty log `max_reorg_depth` and a high number of new generation blocks. 
So at this point I'm not sure if this isn't actually already the correct error message for this scenario. The other part of this PR is to avoid backward syncing to the genesis block. Syncing forward on a new fork is already handled by this part of the implementation.

weilbith:
> I guess he tries to find the parent for the genesis block here.

I analyzed your scenario @schmir and my guess was wrong. 
A lot of new blocks get sealed, which are unknown by the monitoring service and must be fetched. Due to the low `max_reorg_depth` value that is used for the tests, the fetcher switch to sync forward until the last safe block. This fails, because we have a new fork. But since we are in forward sync mode, it gets not detected by this PR.
I think this is a different scenario which is actually already caught by exactly the error message you received. Your sync forward on a fork and that becomes recognized. This happens due to the bad configuration. The purpose of the PR in my thoughts is to avoid the backward syncing to the genesis block.

schmir:
> I think this is a different scenario which is actually already caught by exactly the error message you received. Your sync forward on a fork and that becomes recognized. This happens due to the bad configuration. The purpose of the PR in my thoughts is to avoid the backward syncing to the genesis block.

Does syncing bail out with the same error message if we don't have a fork? If it does, it's a new bug. If it does not, we fail to detect the fork over the initial block, which is what we're trying to do in this PR.

weilbith:
> Does syncing bail out with the same error message if we don't have a fork?

I don't see when this should ever happen else. It is exactly the scenario for a reorg an on a block before the initial one. Else the the parent hash is known known for each new fork. As last option the implementation of the syncing could be incorrect. This are the options I see when this error can occur. So actually this scenario was already solved before the PR.

> If it does not, we fail to detect the fork over the initial block, ...

As I said in fact it is already what we need to detect reorgs when the monitoring service decides to sync forward (e.g. after stopped and restarted after a long pause) as well as when backward (which is the new stuff).

Having this said, I would say we should think about how to refactor both exception messages. Possibly think about a new shared exception type?

weilbith:
Note:
@schmir and I agreed, that the current exception on forward syncing is not specific enough and that additional checks in this scenario must make sure that we really have a fork here.
Furthermore should both scenarios (backward and forward) raise the same type of exception. This allows to combine both scenarios into one test case scenario with a lot of parametrization.

weilbith:
I adjusted the code related to our recent meetings how to handle recognized forks on backward **and** forward synchronization. I extended the test cases agreed and hope the documentation is enough for this complex situations. Please let me know if the changes are satisfying.

@jannikluhn I removed the obsolete function `_should_sync_forwards`. It wasn't used anywhere. assuming it has been just forgot about, I deleted it. Tests are still running as well.

tests/test_fetcher.py
```diff
@@ -127,6 +128,125 @@ def test_noticed_reorg(w3, eth_tester, block_fetcher, report_callback):
     )
 
 
+@pytest.mark.parametrize(
+    "max_reorg_depth, number_of_blocks_on_fork", [(1, 3), (1, 20), (5, 11), (5, 20)]
```
jannikluhn: Can we add the edge case here? Should be `(1, 2)` I guess

tests/test_fetcher.py
```diff
@@ -127,6 +128,125 @@ def test_noticed_reorg(w3, eth_tester, block_fetcher, report_callback):
...
+    branch.
+    The parametrization describes constellation in which the block fetcher will
+    synchronize forwards after the chain has forked.
+    """
+
+    # Ensure that parameter lead to forward syncing.
+    assert number_of_blocks_on_fork > max_reorg_depth * 2
+
+    block_fetcher.max_reorg_depth = max_reorg_depth
+    block_fetcher.initial_block_resolver = ResolveGenesisBlock()
```
jannikluhn: not too happy about these two lines as they feel more like internals. But I guess it's fine for tests.

tests/test_fetcher.py
```diff
@@ -127,6 +128,125 @@ def test_noticed_reorg(w3, eth_tester, block_fetcher, report_callback):
...
+    blockchain becomes reorganized. The block fetcher is idle for that many
+    blocks on the new fork, that forward syncing will be used. This should fail
+    due to fetching blocks from a fork without any hash pointer to the old
+    branch.
+    The parametrization describes constellation in which the block fetcher will
+    synchronize forwards after the chain has forked.
+    """
+
+    # Ensure that parameter lead to forward syncing.
+    assert number_of_blocks_on_fork > max_reorg_depth * 2
```
jannikluhn: this doesn't sound correct, why the `* 2`?

tests/test_fetcher.py
```diff
@@ -127,6 +128,125 @@ def test_noticed_reorg(w3, eth_tester, block_fetcher, report_callback):
     )
 
 
+@pytest.mark.parametrize(
+    "max_reorg_depth, number_of_blocks_on_fork", [(1, 3), (1, 20), (5, 11), (5, 20)]
+)
+def test_fail_on_fetching_fork_when_sync_forwards(
```
jannikluhn: This tests a related, but slightly different scenario from what it should: It's a reorg of depth 1 with a lazy block fetcher that missed a lot of blocks. Here's what I'd expect it to do:

- snapshot
- mine `number_of_blocks_on_fork`, and fetch at the end and in between if necessary (maybe easiest after each single block)
- revert to snapshot
- mine a single block
- fetch and assert the exception

(btw, the test shows that we do have a bug in the code, as this may happen if the node goes offline. But it's unrelated to this PR)

tests/test_fetcher.py
```diff
@@ -127,6 +128,125 @@ def test_noticed_reorg(w3, eth_tester, block_fetcher, report_callback):
...
+    eth_tester.revert_to_snapshot(fork_snapshot_id)
+    eth_tester.mine_blocks(number_of_blocks_on_fork, coinbase=coinbase2)
+
+    #  Try to fetch blocks based on the new fork.
+    with pytest.raises(FetchingForkWithUnkownBaseError):
+        block_fetcher.fetch_and_insert_new_blocks()
+
+
+@pytest.mark.parametrize(
+    "max_reorg_depth, initial_block_number, number_of_blocks_before_initial_height_to_fork, number_of_blocks_on_fork_after_initial_block",
```
jannikluhn: this is a very long line, consider splitting it up like this: `("max_reorg_depth", "initial_block_number", ..)`

tests/test_fetcher.py
```diff
@@ -127,6 +128,125 @@ def test_noticed_reorg(w3, eth_tester, block_fetcher, report_callback):
...
+        (3, 10, 1, 2),
+        # Maximum number_of_blocks_before_initial_height_to_fork related to initial_block_number
+        (3, 10, 10, 2),
+        # Minimum number_of_blocks_on_fork_after_initial_block
+        (3, 10, 2, 1),
+        # Maximum number_of_blocks_on_fork_after_initial_block related related to max_reorg_path
+        (3, 10, 2, 6),
+    ],
+)
+def test_fail_on_fetching_fork_based_on_block_before_inital_one_when_sync_backwards(
```
jannikluhn: See comment on test above

tests/test_fetcher.py
weilbith: This is the edge case. You need `2 * max_reorg_depth +1` blocks.
I just tested it and also adjusted the `assert` statement (just to make it working), but the test failed.
Why do you think this is wrong how it is? I think I'm missing something.

tests/test_fetcher.py
weilbith: To be honest I though how it is better. The alternative would be to replicate the instantiation of `BlockFetcher` like in the used fixture. This is a bit more bulky. So I went for the way of using the block fetcher and adjust it how it is needed only for this case.
But I'm totally fine with doing it the other way. Or is there possibly a third option?

tests/test_fetcher.py
weilbith: Linking this with the below comment of the `assert` statement, I will check this again.

tests/test_fetcher.py
weilbith: I think your scenario is not working. Being in the last step of your procedure. The `fetch_forward_sync_target` would be `0`, so `max_forward_sync_blocks` is `0` as well. Therefore the fetcher will not sync forwards.
That's why my idea is to have a single block on the first branch. After the fork, the fetcher will check only by block numbers if he should synchronize forwards (`max(self.w3.eth.blockNumber - self.max_reorg_depth, 0)`). To have number `> 0`, I need a number of blocks on the fork that is higher than the `max_reorg_depth` (assuming we fork on genesis block, else the offset must be added). If this is given and the fork has as many blocks, the fetcher will sync forwards on the new fork from the old head (here `1 + 1`).
Did this makes it clear? You can call me if I'm totally wrong here.

tests/test_fetcher.py
weilbith: Thank you very much indeed, wasn't aware it would work like this. Much better now!! :+1:

monitor/block_fetcher.py
```diff
@@ -86,10 +101,14 @@ def _insert_branch(self, blocks):
 
     def _insert_first_block(self):
         resolver = self.initial_block_resolver or blocksel.ResolveGenesisBlock()
-        block = resolver.resolve_block(self.w3)
+        block = (
+            self.w3.eth.getBlock(self.initial_blocknr)
+            if self.initial_blocknr
+            else resolver.resolve_block(self.w3)
+        )
```
schmir: Why do we need to look at `self.initial_blocknr` here. We did call `resolve_block` unconditionally before and I fail to see a reason to not do that. How do we hit the path here where `self.initial_blocknr` is truthy?

tests/test_fetcher.py
weilbith: Okay, breaking it down, the block fetcher will sync forwards if `self.w3.eth.blockNumber - self.max_reorg_depth - self.head.number > 0`. Since we mine on the first branch only one block, `self.head.number = 1`. Furthermore we can face `self.w3.eth.blockNumber = number_of_blocks_on_fork` and `max_reorg_depth` is clear. Now we have `max_reorg_depth + 1 = number_of_blocks_on_fork`. So the correct `assert` statement and therefore also the parameters must be `assert number_of_blocks_on_fork > max_reorg_depth + 1`.
I update both. Thanks for asking!!!!!!

tests/test_fetcher.py
weilbith: See the first conversation of this review.

monitor/block_fetcher.py
weilbith: Completely true. Doens't make any sense here. Just bullshit. Remove it immediately.

tests/test_fetcher.py
weilbith: I applied the relation of `max_reorg_depth` and `number_of_blocks_on_fork_after_initial_block` as in the test above.
Else I only see the `BlockFetcher` initialization to adopt here as well. Rest is "special" to the other case I think.

tests/test_fetcher.py
weilbith: Ralf agreed that this approach is fine for the scenario.

tests/test_fetcher.py
jannikluhn: Alright, but the main point is that the current test does not handle the scenario we (or at least I) had in mind when we were discussing it. It's a reorg of just depth 1, not `> max_reorg_depth`.

weilbith:
@schmir I have done so and also added a test case for that. All other block resolvers seem to be safe. I'm not sure if having a safe `ResolveBlockByNumber` safe as well (fall back to genesis or latest block) is better What do you think?

tests/test_fetcher.py
weilbith: This conversation is open since your suggested edge case `(1, 2)` does not work. I hope that the explanation is good and that the changes therefore. Close if you are satisfied.

tests/test_fetcher.py
weilbith: Please let me know if you understood the flow I explained and why I think it must be like this. Just close the conversation it if all is fine.

tests/test_fetcher.py
weilbith: Can be closed since the initialization has been declined.

tests/test_fetcher.py
jannikluhn: Yes, I do understand what you're saying, but I don't see how this is relevant. We want to test the case of a reorg greater than max depth. This test tests a reorg of depth 1.

I see why the scenario I suggested above doesn't work. But it should be easy to fix by just mining more blocks at the end.

weilbith: Sorry, when writing the last comment, your last one wasn't be loaded.
Okay, so let me try to phrase it again before I start. So you want parametrize the number of blocks between the block the fork is based on and the initial block the fetcher syncs from. Afterwards I should just make sure that the fetcher will synchronize forward without any parametrization. So more like the backward test. Is this correct?

weilbith: Wait. We want to test forks before and after the initial block. Correct?

jannikluhn: I thought here we don't care about the initial block at all. This is just "reorgs that are too long result in an error", no?

weilbith: I adjusted the task related to our talk. It is written in a way that it could be easily adopted to future changes when reorgs with less blocks than the `max_reorg_depth` is safe (which is not the case right now).

schmir:
> I'm not sure if having a safe ResolveBlockByNumber safe as well (fall back to genesis or latest block) is better What do you think?

I'm not sure what you mean here.

weilbith:
Just something like `blocknr = min(blocknr, web3.eth.blockNumber)`. But thinking about it twice, I think aborting is the better solution to make the user clear that this does not exist, instead of just selecting any other block.

schmir:
I've pushed another change, which doesn't change the behaviour much, but makes the code a bit cleaner. 
Otherwise I'm now fine with the current state, though I do have to admit that I didn't look into the latest tests.

Please review.

I've said it to andreas yesterday and I'll repeat it here: I would have solved that in like 15 minutes and another 15 minutes if I accidentally introduced a bug some years ago when we didn't have this whole code review process in place.

PR #22, opened 2019-04-10
Prepare dockerhub push

schmir:
This upgrades pip and co and adds a --version command line argument, which we need to tag the docker images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@
 FROM ubuntu:18.04 as builder
 # python needs LANG
 ENV LANG C.UTF-8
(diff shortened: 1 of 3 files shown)


schmir:
setuptools_scm uses the output of `git describe --tags` to compute the version. It doesn't use the output as is, but rather increases the minor version. For the relay server it looks like:
```
~/trustlines/protocol/relay (git)-[master] % git describe --tags
0.6.1-5-gf54cbbc
~/trustlines/protocol/relay (git)-[master] % python setup.py --version
0.6.2.dev5+gf54cbbc.d20190410
```
The date in latter is there because I do have local uncommitted changes.

Issue #29, opened 2019-05-07
Cache current validator set


cducrest:
Involve a second person to discuss on how to do this.
Don't store this cache in the state.

jannikluhn:
Draft: https://gist.github.com/jannikluhn/1c0d79c118a744e9dfa3b21b976fd6a8

cducrest:
done as of https://github.com/trustlines-protocol/tlbc-monitor/pull/30

Issue #28, opened 2019-05-07
Determine current validator set with validator contract


cducrest:
Use events on validator contract to find current set, look at last event before block number to get the set.
related: https://github.com/trustlines-network/project/issues/429
https://github.com/trustlines-protocol/blockchain/issues/62

berndbohmeier:
Done

PR #30, opened 2019-05-08
Implement foundations of validator set fetching

jannikluhn:
WIP as tests are missing, but I'm quite happy with this so far.

Consists of two isolated components that still need to be connected somehow:

- `EpochFetcher`: Gets the chain spec (in the form of a list of `ValidatorDefinitionRange`s) and fetches epochs from the contracts defined there. Remembers where it left off.
- `PrimaryOracle`: Gets `Epoch`s (either static from the chain spec file or fetched from the `EpochFetcher`) and returns the validator for a specific block.


monitor/validators.py
```diff
@@ -33,28 +40,93 @@ def validate_validator_definition(validator_definition):
...
-            start_block_number
-            for start_block_number in descending_start_block_numbers
-            if step >= start_block_number
+    def _fetch_new_epochs_from_contract(self, contract_address: bytes) -> List[Epoch]:
+        contract = self._w3.eth.contract(
+            address=contract_address, abi=VALIDATOR_CONTRACT_ABI
         )
-        current_validator_list = multi_list[start_block_number]
-        return current_validator_list[step % len(current_validator_list)]
+        epoch_start_heights = contract.call().getEpochStartHeights()
```
weilbith: Would you be fine to extend this for the current ABI of #74 or should the contract provide such function? This also affects the fictional `getValidators` function (which name is actually already taken by the Parity ABI for this contract, so must be renamed). Right know the array of heights and the mapping to the address lists are public and you can query their lengths. Afterwards the default ABI functions for public contract variables can be used. But we could make this more comfortable.

monitor/validators.py
```diff
@@ -33,28 +40,93 @@ def validate_validator_definition(validator_definition):
...
 
-def make_primary_function(validator_definition):
-    validate_validator_definition(validator_definition)
+class Epoch(NamedTuple):
+    start_height: int
+    validators: List[bytes]
+
+
+class ValidatorDefinitionRange(NamedTuple):
+    transition_to_height: int
```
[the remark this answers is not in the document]
weilbith: It took me a while to understand the rest of the code due to this. I removed a lot of other comments, because I mixed both values. :see_no_evil:

monitor/validators.py
```diff
@@ -33,28 +40,93 @@ def validate_validator_definition(validator_definition):
...
+    transition_from_height: int
```
weilbith: Assuming this is the "end" of the range, would it be useful to set max int as default? Because I guess that's what is necessary for the rest of the code for the last defined range in the specification.

weilbith:
This related to #28 and #29, correct?

monitor/validators.py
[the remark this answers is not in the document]
jannikluhn: For a chain spec like this:

```
{
    0: [a, b],
    10: [c, d]
}
```
the `transition_to_height` would be 0, the `transition_from_height` would be 10 for the validator set `[a, b]`.

I agree that the naming is a bit hard to grasp on first read, but then it should be quite clear. Do you have a better suggestion?

monitor/validators.py
```diff
@@ -33,28 +40,93 @@ def validate_validator_definition(validator_definition):
...
-    ].items():
-        start_block_number = int(start_block_number_str)
-        addresses = [
-            to_canonical_address(address) for address in multi_list_entry["list"]
+    def _get_validators(self, block_height: int) -> List[bytes]:
+        started_epochs = [
+            epoch for epoch in self._epochs if epoch.start_height < block_height
         ]
-        multi_list[start_block_number] = addresses
+        return started_epochs[-1].validators
```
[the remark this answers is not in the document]
jannikluhn: Fixed this morning, will raise a `ValueError`.

monitor/validators.py
jannikluhn: Yep, it's just a placeholder for now. I'll use whatever API will be provided by the contract.

Doesn't Solidity support function overloading though?

monitor/validators.py
jannikluhn: Hmm, you're right, this case isn't handled yet. I'd just set it to `None` though as Python doesn't have a max int.

monitor/validators.py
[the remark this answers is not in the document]
weilbith: Yeah, I guess now I'm also into it. Possible a comment would help each one reading it for the first time?
In your example, is `transition_from_height` not `9`? :thinking:

monitor/validators.py
weilbith: Okay, so I'm about to improve the interface.
Haven't used overloading in Solidity so far. But a first test it compiles. So I would go for it. :+1:

monitor/validators.py
weilbith: But for None, the `>` will not work, right?

monitor/validators.py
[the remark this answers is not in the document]
jannikluhn: No, it's a single transition at block 10.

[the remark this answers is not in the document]
weilbith: Okay, I guess in the end it is a [`>`, `<`] vs [`>=`, '<=`] for the code, so I'm fine.
Please consider to add a comment, else all good.

monitor/validators.py
schmir-at-bb: You can use `math.inf` as a guard!

jannikluhn:
This is mostly done and ready for a preliminary review. It will have to be rebased on top of #31 once it is merged.

Only thing missing is unit tests for the epoch fetcher which shouldn't affect the rest of the code too much. Might take a bit though as I have to find a good way to import the validator set contract from another repo.

tests/test_primary_oracle.py
```diff
@@ -0,0 +1,71 @@
...
+
+def test_get_primary_no_epochs():
+    primary_oracle = PrimaryOracle()
+    for height in range(5):
+        for step in range(10):
+            with pytest.raises(ValueError):
+                primary_oracle.get_primary(height, step)
+
+
+def test_get_primary_late_epoch():
```
weilbith: Since we test this, should we not alternatively make sure within the code, that the epochs are chained? So the transition from and to fit in the sorted list? Then we could also require that the first entry must start at `0`.

monitor/main.py
```diff
@@ -114,35 +115,45 @@ def __init__(
...
+            head_number=self.block_fetcher.head.number,
+        )
+
+        if number_of_new_blocks == 0:
+            time.sleep(BLOCK_FETCH_INTERVAL)
+
+    def _update_epochs(self) -> None:
+        new_epochs = self.epoch_fetcher.fetch_new_epochs()
+        for epoch in new_epochs:
+            self.primary_oracle.add_epoch(epoch)
```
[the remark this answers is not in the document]
jannikluhn: I guess only for tests. However, I don't like to add two methods that do basically the same thing.

tests/test_primary_oracle.py
```diff
@@ -0,0 +1,71 @@
...
+    primary_oracle = PrimaryOracle()
+    primary_oracle.add_epoch(Epoch(0, [VAL1, VAL2]))
+    for height in range(10):
+        for step in range(0, 10, 2):
+            assert primary_oracle.get_primary(height, step) == VAL1
+        for step in range(1, 10, 2):
+            assert primary_oracle.get_primary(height, step) == VAL2
+
+
+def test_get_primary_two_epochs():
```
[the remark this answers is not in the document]
jannikluhn: I'd rather not, I think it's clearer as it is now.

tests/test_primary_oracle.py
jannikluhn: `Epoch` only has a start, not an end. That's because it's hard to know when they end (and not really necessary for anything).

I don't see a good reason to enforce this. At some point maybe we don't look at past epochs at all and then we'd have to change it back.

weilbith: Good arguments.

monitor/main.py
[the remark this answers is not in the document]
weilbith: > I don't like to add two methods

No my idea was just expect an array as input and do the work at this single place, instead of require this annoying loop at every usage. Even if then occur some cases you want to insert a single epoch, you can just put it into parenthesis on call. :man_shrugging:

jannikluhn:
I realized there are some issues with this, I'm still working on it.

monitor/validators.py
```diff
@@ -93,89 +92,189 @@ def get_primary(self, *, height: int, step: int):
...
+            if epoch.start_height not in self._epochs:
+                bisect.insort(self._ordered_start_heights, epoch.start_height)
+            self._epochs[epoch.start_height] = epoch
 
-        if epoch.start_height in self._epoch_start_heights:
-            raise ValueError(
-                "Another non-static epoch starting at this height already exists"
+            self._remove_epochs_rendered_obsolete(epoch)
+
+    def _is_relevant(self, epoch: Epoch) -> bool:
```
weilbith: It took me a while to find out what type of relevance you are checking here. So a comment could be useful.
Just that we are on the same page: you ignore `Epoch`s which are derived from an earlier `ValidatorDefinitionRange` index but a higher start height. That sound also confusing when I'm writing it.
Anyways I'm not sure if this is correct. The multi-set definition in the chain specification is not a list and becomes sorted by transition block numbers (I tested this). But probably I just don't got the purpose.

monitor/validators.py
```diff
@@ -33,28 +41,240 @@ def validate_validator_definition(validator_definition):
...
+        self._w3 = w3
+        self._contract = w3.eth.contract(
+            address=validator_definition_range.contract_address,
+            abi=VALIDATOR_CONTRACT_ABI,
+        )
+
+        self._transition_to_height = validator_definition_range.transition_to_height
+        self._validator_definition_index = validator_definition_index
+
+        self._last_fetch_height: Optional[int] = None
```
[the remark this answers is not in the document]
jannikluhn: The height is not fetched. This is the time a fetch happened.

monitor/validators.py
```diff
@@ -33,28 +41,240 @@ def validate_validator_definition(validator_definition):
...
-    multi_list = {}
-    for start_block_number_str, multi_list_entry in validator_definition[
-        "multi"
-    ].items():
-        start_block_number = int(start_block_number_str)
-        addresses = [
-            to_canonical_address(address) for address in multi_list_entry["list"]
+        # epoch start heights will only be updated in finalizeChange which is called at most once
+        # per block
+        assert len(set(epoch_start_heights)) == len(epoch_start_heights)
```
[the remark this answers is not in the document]
jannikluhn: I think if the contract is broken it doesn't really matter how the monitor behaves, does it?

monitor/validators.py
```diff
@@ -93,89 +92,189 @@ def get_primary(self, *, height: int, step: int):
...
+            self._latest_fetched_epoch = new_epochs[-1]
+
+        return new_epochs
+
+
+class EpochFetcher:
+    def __init__(
+        self, w3: Web3, validator_definition_ranges: List[ValidatorDefinitionRange]
+    ) -> None:
+        self._w3 = w3
```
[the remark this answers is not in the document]
jannikluhn: yes, can be removed

monitor/validators.py
```diff
@@ -93,89 +92,189 @@ def get_primary(self, *, height: int, step: int):
...
+
+        self._remove_stale_fetchers()
+
         return new_epochs
+
+    def _remove_stale_fetchers(self) -> None:
+        while self._pop_first_fetcher_if_stale():
+            continue
+
+    def _pop_first_fetcher_if_stale(self) -> Optional[ContractEpochFetcher]:
```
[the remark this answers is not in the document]
jannikluhn: Basically once a epoch from the second contract has started, we never have to check the first contract again. Doesn't this make sense?

>  Moreover we assume that the given validator_definition_ranges are sorted, right? Possibly we should secure this with an assert? Else the rest of the logic fails and remove them accidentally.

Yes, I'll add a check in `__init__` for this.

monitor/validators.py
jannikluhn: The original idea was to prevent an epoch from an old range to overwrite one from a new range. This might not happen in practice though, depending on how finalization works. But I think it's better to have something like this anyways. It assumes that the validator definition ranges are ordered. In this case, it should be correct, no?

tests/test_epoch_fetcher.py
```diff
@@ -31,4 +93,216 @@ def test_get_static_epochs():
...
+    w3, tester, validator_set_contract
+):
+    val_def, (contract,) = initialize_scenario(
+        validator_set_contract, transition_heights=[100]
+    )
+    fetcher = ContractEpochFetcher(w3, val_def[0], 0)
+
+    mine_until(w3, tester, 123)
+    fetcher.fetch_new_epochs()
+    assert fetcher.last_fetch_height == 123
```
weilbith: I would argue if you expect this test to change the `last_fetch_height`, you should make sure that it is `None` in the beginning. Decide if you think the same and want to continue this for the next test cases.

monitor/validators.py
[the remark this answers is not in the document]
weilbith: I just understood this on my way home yesterday. Makes totally sense yes. :+1:

monitor/validators.py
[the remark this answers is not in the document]
weilbith: Hmm. Probably this is okay. I my thoughts the contract should still work with a wrong history, while the monitor would not.

monitor/validators.py
weilbith: Hmm. I also can't imagine a scenario where this would happen. So it is hard to verify. But since it secures the correct chain of epochs, I would argue this check is not useless. :thinking:

monitor/validators.py
```diff
@@ -1,14 +1,112 @@
...
 
-from eth_utils import is_hex_address, last
-
-
-VALIDATOR_CONTRACT_ABI = None
+from eth_utils import is_hex_address, decode_hex
+from eth_utils.toolz import last
+
+
+VALIDATOR_CONTRACT_ABI = [
```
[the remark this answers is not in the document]
jannikluhn: I'd rather not change the ABI, but I can put it in a separate file.

tests/test_epoch_fetcher.py
jannikluhn: this is checked in `test_contract_epoch_fetcher_initialization`

weilbith: Hmm. Yes, but... Okay, accepted.

monitor/validators.py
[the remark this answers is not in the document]
weilbith: If you feel better then. :man_shrugging:

tests/test_epoch_fetcher.py
```diff
@@ -0,0 +1,308 @@
...
+            contract_address=None,
+            validators=[VALIDATOR2],
+        ),
+    ]
+    epochs = get_static_epochs(ranges)
+    assert epochs == [Epoch(0, [VALIDATOR1], 0), Epoch(10, [VALIDATOR2], 2)]
+
+
+def initialize_scenario(
+    validator_set_contract: Contract, transition_heights: Sequence[int] = None
```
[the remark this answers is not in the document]
jannikluhn: Why? As the number of parameters is quite small and there's no possibility for confusion I think it's fine to not enforce this.

monitor/skip_reporter.py
```diff
@@ -62,7 +66,9 @@ def __call__(self, block):
         # report misses
         missed_steps = self.get_missed_steps()
         for step in missed_steps:
-            primary = self.get_primary_for_step(step)
+            primary = self.primary_oracle.get_primary(
+                height=2 ** 64 - 1, step=step
```
berndbohmeier: Comment what needs to be fixed, and what this constant is.

monitor/main.py
```diff
@@ -114,35 +120,45 @@ def __init__(
...
-                )
-
-                if number_of_new_blocks == 0:
-                    time.sleep(BLOCK_FETCH_INTERVAL)
+                self._run_cycle()
         finally:
             self.skip_file.close()
 
+    def _run_cycle(self) -> None:
+        self._update_epochs()
```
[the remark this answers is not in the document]
jannikluhn: Hm, you're right, it could.

To fix this, we'd have to

1) store the maximum block height until which all epochs are definitely known in the primary oracle
2) fail if we call `get_primary` with a height larger than this
3) handle this error in the skip and offline reporters in some way

3) seems hard to do in a good way. We'd probably get pretty much only false positives as the validator set will change almost never anyways. So I'm inlined to leaving it like this, unless you disagree or have a simple way to fix this.

monitor/validators.py
```diff
@@ -33,28 +139,242 @@ def validate_validator_definition(validator_definition):
...
-        start_block_number = int(start_block_number_str)
-        addresses = [
-            to_canonical_address(address) for address in multi_list_entry["list"]
-        ]
-        multi_list[start_block_number] = addresses
 
-    descending_start_block_numbers = sorted(multi_list.keys(), reverse=True)
+class ValidatorDefinitionRange(NamedTuple):
+    # first block height at which this definition is considered
+    transition_to_height: int
```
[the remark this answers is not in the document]
jannikluhn: I originally didn't like start height as we have the same name for epochs and I thought it could be confusing. But now I don't really care anymore. Will rename it once I rebased it on top of #31.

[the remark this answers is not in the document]
weilbith: @jannikluhn didn't we discussed this with @saschagoebel as well, since he and I also were confused, and got some new names? :thinking:

monitor/main.py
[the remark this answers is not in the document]
berndbohmeier: does not matter, can be changed.

monitor/main.py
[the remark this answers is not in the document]
berndbohmeier: 3. We could also instead give `fetch_and_insert_blocks` a max blocknumber parameter and it would check to not fetch beyond that limit.

monitor/validators.py
```diff
@@ -56,9 +57,33 @@ class ValidatorDefinitionRange(NamedTuple):
     validators: Optional[List[bytes]]
 
 
+def get_static_epochs(
```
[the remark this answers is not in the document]
berndbohmeier: Should be fine.

monitor/validators.py
```diff
@@ -112,28 +128,281 @@ def get_validator_definition_ranges(validator_definition):
...
+        self._contract_epoch_fetchers = [
+            ContractEpochFetcher(w3, validator_definition_range, index)
+            for index, validator_definition_range in enumerate(
+                self._validator_definition_ranges
+            )
+            if validator_definition_range.is_contract
+        ]
+
+    @property
+    def max_height(self) -> int:
```
berndbohmeier: This should be also named `last_fetched_height`, otherwise I would expect it to check what it right now is the max height it can know.

monitor/validators.py
```diff
@@ -112,28 +128,281 @@ def get_validator_definition_ranges(validator_definition):
...
+    def latest_fetched_epoch(self) -> Optional[Epoch]:
+        return self._latest_fetched_epoch
+
+    @property
+    def last_fetch_height(self) -> Optional[int]:
+        return self._last_fetch_height
+
+    def fetch_new_epochs(self) -> List[Epoch]:
+        self._last_fetch_height = self._w3.eth.blockNumber
+        epoch_start_heights = self._contract.call().getEpochStartHeights()
```
berndbohmeier: `call()` is deprecated, please use `self._contract.functions.getEpochStartHeights().call()`

Please check your setup, why you did not see the deprecated warnings in the console.

tests/test_epoch_fetcher.py
```diff
@@ -1,12 +1,74 @@
...
+
+
+@pytest.fixture
+def w3(tester):
+    provider = EthereumTesterProvider(tester)
+    return Web3(provider)
+
+
+@pytest.fixture
+def contracts_json():
```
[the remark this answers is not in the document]
berndbohmeier: This here can not be solved, because deploy tools expect the contracts to be there to get compiled. 

But the other stuff like w3, tester are all provided by the plugin.

monitor/skip_reporter.py
```diff
@@ -38,57 +43,64 @@ def from_fresh_state(cls, *args, **kwargs):
...
+    def remove_open_skipped_proposals_with_step(self, step):
+
+        self.open_skipped_proposals = {
+            proposal for proposal in self.open_skipped_proposals if proposal.step != step
+        }
+
+    def update_open_skipped_proposals(self, current_step, latest_block_height):
         if current_step > self.latest_step:
-            self.open_steps |= set(range(self.latest_step + 1, current_step + 1))
+            for step in range(self.latest_step + 1, current_step + 1):
```
berndbohmeier: Why did you change that into a loop, I think we should have kept using set comprehension

berndbohmeier: Here is actually a off by one error. This should not include current_step

cducrest: We will keep it this way.

cducrest: I changed it

tests/test_epoch_fetcher.py
```diff
@@ -0,0 +1,322 @@
...
+    validators = initialize_validators(contract)
+
+    fetcher = ContractEpochFetcher(w3, val_def[0], 0)
+
+    epochs = fetcher.fetch_new_epochs()
+    assert epochs == [Epoch(100, validators, 0)]
+
+
+def test_fetch_second_time(w3, validator_set_contract):
+    val_def, (contract,) = initialize_scenario(
```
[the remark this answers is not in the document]
berndbohmeier: I decided that this is not urgent.

monitor/validators.py
```diff
@@ -173,7 +175,7 @@ def get_primary(self, *, height: int, step: int):
         index = step % len(validators)
         return validators[index]
 
-    def _get_validators(self, block_height: int) -> List[bytes]:
+    def _get_validators(self, block_height: int) -> Sequence[bytes]:
```
jannikluhn: I don't care too much, but I thought using `List` here is preferable to `Sequence` since return types should be as concrete as possible.

monitor/skip_reporter.py
```diff
@@ -38,57 +43,64 @@ def from_fresh_state(cls, *args, **kwargs):
...
     @staticmethod
     def get_fresh_state():
-        return SkipReporterState(latest_step=None, open_steps=set())
+        return SkipReporterState(latest_step=None, open_skipped_proposals=set())
 
     @property
     def state(self):
         return SkipReporterState(
-            latest_step=self.latest_step, open_steps=self.open_steps
+            latest_step=self.latest_step, open_skipped_proposals=self.open_skipped_proposals  # TODO: is it ok that I store a set of SkippedProposals instead of ints
```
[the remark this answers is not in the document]
jannikluhn: Ah, Github shows this as outdated. I removed some commits independently because I can't properly shift-click  apparently.

monitor/skip_reporter.py
```diff
@@ -38,57 +43,64 @@ def from_fresh_state(cls, *args, **kwargs):
...
+                self.open_skipped_proposals.remove(proposal)
 
-    def update_open_steps(self, current_step):
+    def update_open_skipped_proposals(self, current_step, latest_block_height):
         if current_step > self.latest_step:
-            self.open_steps |= set(range(self.latest_step + 1, current_step + 1))
+            for step in range(self.latest_step + 1, current_step + 1):
+                skipped = SkippedProposal(current_step, latest_block_height)
+                self.open_skipped_proposals.add(skipped)
+            # self.open_steps |= set(range(self.latest_step + 1, current_step + 1))
```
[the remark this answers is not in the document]
jannikluhn: same

monitor/offline_reporter.py
```diff
@@ -99,26 +100,42 @@ def _clear_old_skips(self, current_step):
             },
         )
 
-    def _is_offline(self, validator, skips, current_step):
+    def _is_offline(self, validator, recent_skipped_proposals, skipped_proposal):
 
-        assigned_steps = self._get_assigned_steps_in_offline_window(validator, current_step)
+        if skipped_proposal.step - self.offline_window_size <= 0:
```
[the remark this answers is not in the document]
berndbohmeier: The problem is that at the start we have not enough blocks, so the following might try to check blocks before the genesis.

monitor/validators.py
berndbohmeier: Why should they be as concrete as possible?

jannikluhn: To give the user as much information about what we return as possible. For example, if they want to use some specific list method, they can't do so if we use `Sequence`.

monitor/offline_reporter.py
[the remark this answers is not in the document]
jannikluhn: This could be handled gracefully, e.g. by saying that no one is assigned to the genesis block or earlier). But I'm also okay with keeping this check (but see the first comment).

monitor/offline_reporter.py
```diff
@@ -99,26 +100,42 @@ def _clear_old_skips(self, current_step):
...
-            if self.primary_oracle.get_primary(height=block_height, step=step) == validator:
+        for step in window:
+            if (
+                self.primary_oracle.get_primary(height=block_height_of_step, step=step)
+                == validator
+            ):
                 assigned_steps += 1
             if step not in self.missed_steps:
-                block_height -= 1
+                block_height_of_step -= 1
```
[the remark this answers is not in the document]
cducrest: can also move 
`if primary in self.reported_validators: return  # ignore validators that have already been reported`

in `call` to a few lines below so that is is correct.

monitor/offline_reporter.py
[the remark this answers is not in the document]
berndbohmeier: okay.

monitor/offline_reporter.py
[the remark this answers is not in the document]
jannikluhn: yes, this could work as well. I don't like how it takes and changes state from so many sources, but I guess that's just inherent in the design of the monitor by now.

monitor/validators.py
berndbohmeier: ah ok, but you would agree that for the input we should accept the least concrete?

jannikluhn: exactly, for inputs it should be the other way around

monitor/offline_reporter.py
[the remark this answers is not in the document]
berndbohmeier: We decided to go with completely different approach.

monitor/offline_reporter.py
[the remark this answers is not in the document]
berndbohmeier: different approach, see above

monitor/validators.py
berndbohmeier: Agree.

monitor/offline_reporter.py
```diff
@@ -10,10 +10,20 @@
 from monitor.skip_reporter import SkippedProposal
 
 
+class OfflineStep(NamedTuple):
```
jannikluhn: I think `OfflineInterval` (as in the comment below) would be a bit clearer.

monitor/offline_reporter.py
```diff
@@ -10,10 +10,20 @@
...
+    # so length is the number of current validators
+    length: int
+
+
 class OfflineReporterStateV1(NamedTuple):
-    missed_steps: Any
-    reported_validators: Any
-    recent_skips_by_validator: Any
+    reported_validators: Set[bytes]
+    recent_skips_by_validator: Dict[bytes, List[OfflineStep]]
```
jannikluhn: suggesting `recent_offline_step/interval_by_validator` for consistency

monitor/offline_reporter.py
```diff
@@ -73,69 +87,56 @@ def __call__(self, primary, skipped_proposal: SkippedProposal):
...
+            offline_steps = self.recent_offline_steps_by_validator.pop(primary)
 
             for callback in self.report_callbacks:
-                callback(primary, list(sorted(skips)))
+                callback(
+                    primary,
+                    list(sorted(offline_step.step for offline_step in offline_steps)),
+                )
+
+    def _update_skips(self, validator: bytes, skipped_proposal: SkippedProposal):
```
jannikluhn: `_update_offline_intervals` for consistency

monitor/offline_reporter.py
```diff
@@ -10,10 +10,20 @@
...
+
 class OfflineReporterStateV1(NamedTuple):
-    missed_steps: Any
-    reported_validators: Any
-    recent_skips_by_validator: Any
+    reported_validators: Set[bytes]
+    recent_skips_by_validator: Dict[bytes, List[OfflineStep]]
+
+    # offline time in number of steps
+    offline_time_by_validator: Dict[bytes, int]
```
jannikluhn: This doesn't seem to be used in `_is_offline` where it matters, unless I'm missing something.

I can't make up my mind if it's better to keep or get rid of it: On the one hand it's very clear and expressive, on the other it's just duplicated data that needs to be managed separately. So I'm not sure, I guess I'd be ok with either.

berndbohmeier: I forgot to use it, my bad. This was introduced for performance reasons.

monitor/offline_reporter.py
berndbohmeier: Changed

monitor/offline_reporter.py
berndbohmeier: agree

PR #31, opened 2019-05-10
Feature/chainspec parser

### LINKED ISSUE #27: Parse config file to find active contract at specific block number
Need to still support the past and find previous static list / contract
related: https://github.com/trustlines-network/project/issues/429

saschagoebel:
Fixes #27 

Extends valid chainspec syntax and parses config into ValidatorDefinitionRanges as a precondition to #30

### DIFF
--- a/monitor/validators.py
+++ b/monitor/validators.py
@@ -1,8 +1,9 @@
 from collections.abc import Mapping
-from itertools import tee, islice, chain
+from itertools import chain
 from typing import NamedTuple, List, Optional
 
 from eth_utils import is_hex_address, to_canonical_address
+from eth_utils.toolz import sliding_window
 
 
 def validate_validator_definition(validator_definition):
@@ -55,13 +56,15 @@ def validate_validator_definition(validator_definition):
                         "Validator contract address must be a single hex address"
                     )
             else:
-                assert False, "Unreachable. Multi list entry type has already been validated."
+                assert (
+                    False
+                ), "Unreachable. Multi list entry type has already been validated."
 
 
 # Added for compatibility with the upcoming pull request
 class ValidatorDefinitionRange(NamedTuple):
-    transition_to_height: int
-    transition_from_height: int
+    enter_height: int
+    leave_height: int
     is_contract: bool
     contract_address: Optional[bytes] = None
     validators: Optional[List[bytes]] = None
@@ -72,15 +75,17 @@ def get_validator_definition_ranges(validator_definition):
 
     sorted_definition = sorted(
         # Lambda tuple destructuring has been removed from Python 3 (https://www.python.org/dev/peps/pep-3113/) :-(
-        validator_definition["multi"].items(), key=lambda item: int(item[0])
-
+        validator_definition["multi"].items(),
+        key=lambda item: int(item[0]),
     )
 
-    items, nexts = tee(sorted_definition, 2)
-    nexts = chain(islice(nexts, 1, None), [[None, None]])
-
     result = []
-    for (range_height, range_config), (next_range_height, _) in zip(items, nexts):
+
+    # Iterate over all configurations. Add an extra empty item for the sliding window to slide to the very end.
+    # Alternatively we'd have to do some additional processing which would further complicate the code
+    for (range_height, range_config), (next_range_height, _) in sliding_window(
+        2, chain(sorted_definition, [[None, None]])
+    ):
         [(config_type, config_data)] = range_config.items()
 
         validators = None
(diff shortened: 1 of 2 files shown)


monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+                ):
+                    raise ValueError(
+                        "Static validator list must only contain hex addresses"
+                    )
+
+            elif multi_list_entry_type in ["safeContract", "contract"]:
+                if not is_hex_address(multi_list_entry_data):
+                    raise ValueError(
+                        "Dynamic validator list must be a single hex address"
+                    )
```
jannikluhn: Adding a
```Py
else:
    assert False, "unreachable"
```
at the end would be nice to make it clear that we haven't forgot something.

monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+                    not is_hex_address(address) for address in multi_list_entry_data
+                ):
+                    raise ValueError(
+                        "Static validator list must only contain hex addresses"
+                    )
+
+            elif multi_list_entry_type in ["safeContract", "contract"]:
+                if not is_hex_address(multi_list_entry_data):
+                    raise ValueError(
+                        "Dynamic validator list must be a single hex address"
```
jannikluhn: Not sure about the message, how about
```suggestion
                        "Validator contract address must be a single hex address"
```
?

monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+    is_contract: bool
+    contract_address: Optional[bytes] = None
+    validators: Optional[List[bytes]] = None
+
+
+def get_validator_definition_ranges(validator_definition):
+    validate_validator_definition(validator_definition)
+
+    sorted_definition = sorted(
+        validator_definition["multi"].items(), key=lambda kv: int(kv[0])
```
jannikluhn:
```suggestion
        validator_definition["multi"].items(), key=lambda height, _: int(height)
```

monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+    )
+
+    items, nexts = tee(sorted_definition, 2)
+    nexts = chain(islice(nexts, 1, None), [[None, None]])
+
+    result = []
+    for (range_height, range_config), (next_range_height, _) in zip(items, nexts):
```
jannikluhn: I like to use `sliding_window` from the `toolz` library for stuff like this. Can be imported from `eth_utils.toolz`.

monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+        [(config_type, config_data)] = range_config.items()
+
+        if config_type == "list":
```
jannikluhn: I suggest creating the `ValidatorDefinitionRange` only after the if/else block and only extracting the parameters here. So something like 

```
if config_type == "list":
    is_contract = False
    validators = config_data
    contract_address = None
else:
    ...
result.append(ValidatorDefinitionRange(...))
```

Saves some duplication

monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+            result.append(
+                ValidatorDefinitionRange(
+                    transition_from_height=range_height,
+                    transition_to_height=next_range_height,
+                    is_contract=False,
+                    validators=config_data,
+                )
+            )
+        else:
```
jannikluhn: it's probably fine as is, but we could add an `assert config_type in ("contract", "safeContract")` (or put that in an `elif` and add `assert False, "unreachable"` in an `else block`) to make it a bit easier to understand what's going on if one hasn't read the validation function right before.

monitor/validators.py
```diff
@@ -21,16 +23,82 @@ def validate_validator_definition(validator_definition):
...
+    nexts = chain(islice(nexts, 1, None), [[None, None]])
+
+    result = []
+    for (range_height, range_config), (next_range_height, _) in zip(items, nexts):
+        [(config_type, config_data)] = range_config.items()
+
+        if config_type == "list":
+            result.append(
+                ValidatorDefinitionRange(
+                    transition_from_height=range_height,
```
jannikluhn: Oh, sorry, missed that on first review.  This is the wrong way around. `transition_to_height` is supposed to be the height we transition to this definition, `transition_from_height` is the height we go to the next one.

I'm open for better names btw since it also confused @thore if you have a suggestion

Issue #32, opened 2019-05-10
Monitor uses step instead of height to determine validator set


jannikluhn:
The validator set is determined by the block height. Only choosing the primary from the validator set is done by step. Currently, we use the step for both which effectively means we only ever use the last validator set as step is usually much larger than height.

This needs fixes at the following sites:

- https://github.com/trustlines-protocol/tlbc-monitor/blob/master/monitor/validators.py#L36 (note that this will become obsolete once #30 is merged)
- https://github.com/trustlines-protocol/tlbc-monitor/blob/master/monitor/skip_reporter.py#L65
- https://github.com/trustlines-protocol/tlbc-monitor/blob/master/monitor/offline_reporter.py#L91

cducrest:
done as of https://github.com/trustlines-protocol/tlbc-monitor/pull/30

PR #33, opened 2019-05-14
Fix/height to determine validators

cducrest:
Related to: https://github.com/trustlines-protocol/tlbc-monitor/issues/32
I wait for the PR https://github.com/trustlines-protocol/tlbc-monitor/pull/30 to be merged before rebasing / adding tests if necessary


monitor/skip_reporter.py
```diff
@@ -38,54 +43,64 @@ def from_fresh_state(cls, *args, **kwargs):
...
+        for proposal in self.open_skipped_proposals:
+            if proposal.step == step:
+                self.open_skipped_proposals.remove(proposal)
 
-    def update_open_steps(self, current_step):
+    def update_open_skipped_proposals(self, current_step, latest_block_height):
         if current_step > self.latest_step:
-            self.open_steps |= set(range(self.latest_step + 1, current_step + 1))
+            for step in range(self.latest_step + 1, current_step + 1):
+                skipped = SkippedProposal(current_step, latest_block_height)
```
jannikluhn: `skipped` -> `proposal` or `skipped_proposal`

monitor/skip_reporter.py
```diff
@@ -38,54 +43,64 @@ def from_fresh_state(cls, *args, **kwargs):
...
+                self.open_skipped_proposals.add(skipped)
+            # self.open_steps |= set(range(self.latest_step + 1, current_step + 1))
```
jannikluhn: can be removed

monitor/skip_reporter.py
```diff
@@ -38,54 +43,64 @@ def from_fresh_state(cls, *args, **kwargs):
...
-        self.open_steps -= set(missed_steps)
+        self.open_skipped_proposals -= set(reported_proposals)
+
+    def remove_open_skipped_proposals_with_step(self, step):
+        for proposal in self.open_skipped_proposals:
+            if proposal.step == step:
+                self.open_skipped_proposals.remove(proposal)
 
-    def update_open_steps(self, current_step):
+    def update_open_skipped_proposals(self, current_step, latest_block_height):
```
jannikluhn: Not sure about the naming of `latest_block_height` as it's not necessarily the latest one. Maybe `sibling_height` or `replacement_height` or just `height`?

monitor/offline_reporter.py
```diff
@@ -62,10 +67,14 @@ def state(self):
     def register_report_callback(self, callback):
         self.report_callbacks.append(callback)
 
-    def __call__(self, primary, step):
+    def __call__(self, primary, step, latest_block_height):
```
jannikluhn: Same as below, here just `height` seems fine.

monitor/offline_reporter.py
```diff
@@ -91,15 +100,24 @@ def _clear_old_skips(self, current_step):
...
         skip_rate = Fraction(len(missed_steps_in_window), len(assigned_steps))
         return skip_rate > self.allowed_skip_rate
+
+    def _get_assigned_steps_in_offline_window(self, validator, current_step):
+        window = range(current_step - self.offline_window_size + 1, current_step + 1)
+        window = reversed(window)
+        assigned_steps = 0
+        block_height = self.latest_block_height
+        for step in window:
+            if self.primary_oracle.get_primary(height=block_height, step=step):
```
jannikluhn: This is not really correct, right? It could be that the validator set changed during the window, so we would have to update the height in some way. I don't see a good way to do this right now. It would probably be best to call the reporter not only when blocks are skipped, but also when they produce blocks. But this doesn't fit in the current architecture.

monitor/offline_reporter.py
```diff
@@ -91,15 +100,24 @@ def _clear_old_skips(self, current_step):
...
+        assigned_steps = self._get_assigned_steps_in_offline_window(validator, current_step)
+
+        window = range(current_step - self.offline_window_size + 1, current_step + 1)
         missed_steps_in_window = [step for step in skips if step in window]
 
         skip_rate = Fraction(len(missed_steps_in_window), len(assigned_steps))
         return skip_rate > self.allowed_skip_rate
+
+    def _get_assigned_steps_in_offline_window(self, validator, current_step):
+        window = range(current_step - self.offline_window_size + 1, current_step + 1)
```
jannikluhn: `reversed_window` to avoid reusing the same variable for two different things

monitor/skip_reporter.py
cducrest: It is the latest block height the reporter is aware of, is it not?
If it is not the case then my (code) logic is most likely flawed.

monitor/offline_reporter.py
cducrest: The block height is modified three lines below.
I ask for the primary for the current block height, and for every not missed step (so step where a block we produced, I believe) I decrease the `block_height`.
This algorithm is wrong if we do not count as skip when a primary produces a block with the wrong (not increasing) block height. Do we do this?

PR #36, opened 2019-05-16
Add option to specify max block number in fetcher

### LINKED ISSUE #34: Allow block fetcher only to fetch up to a certain block number
This is necessary to allow it to work right together with the `get_primary` function of the PrimaryOracle, because it only can know the primaries reliably up to a certain point.

berndbohmeier:
Add an option to specify a maximum blocknumber until which the fetcher
will fetch. This number is inclusive.

Fix a bug where the first block of the new branch was not counted
towards the fetched blocks. This requiered to adjust some wrong tests.

Closes #34

### DIFF
--- a/monitor/block_fetcher.py
+++ b/monitor/block_fetcher.py
@@ -118,9 +118,9 @@ class BlockFetcher:
         self._insert_branch([block])
 
     def fetch_and_insert_new_blocks(
-        self, *, max_number_of_blocks=5000, max_block_number: int = None
+        self, *, max_number_of_blocks=5000, max_block_height: int = None
     ):
-        """Fetches up to `max_number_of_blocks` blocks and only up to blocknumber `max_block_number` (inclusive)
+        """Fetches up to `max_number_of_blocks` blocks and only up to blocknumber `max_block_height` (inclusive)
         and updates the internal state
             If a full branch is fetched it also inserts the new blocks
             Returns the number of fetched blocks
@@ -136,16 +136,16 @@ class BlockFetcher:
 
         # sync forwards at most up until the forward sync target, but no more than
         # max_number_of_blocks
-        max_forward_block_number = (
+        max_forward_block_height = (
             self.fetch_forward_sync_target()
-            if max_block_number is None
-            else min(self.fetch_forward_sync_target(), max_block_number)
+            if max_block_height is None
+            else min(self.fetch_forward_sync_target(), max_block_height)
         )
         max_forward_sync_blocks = max(0, max_number_of_blocks - number_of_synced_blocks)
 
         number_of_synced_blocks += self._sync_forwards(
             max_number_of_blocks=max_forward_sync_blocks,
-            max_block_number=max_forward_block_number,
+            max_block_height=max_forward_block_height,
         )
 
         # sync backwards until we have synced max_number_of_blocks in total or we are fully synced
@@ -154,7 +154,7 @@ class BlockFetcher:
 
         number_of_synced_blocks += self._sync_backwards(
             max_number_of_blocks=max_backward_sync_blocks,
-            max_block_number=max_block_number,
+            max_block_height=max_block_height,
         )
 
         return number_of_synced_blocks
@@ -163,11 +163,11 @@ class BlockFetcher:
         return max(self.w3.eth.blockNumber - self.max_reorg_depth, 0)
 
     def _sync_forwards(
-        self, *, max_number_of_blocks: int, max_block_number: int
+        self, *, max_number_of_blocks: int, max_block_height: int
     ) -> int:
         block_numbers_to_fetch = range(
             self.head.number + 1,
-            min(self.head.number + 1 + max_number_of_blocks, max_block_number + 1),
+            min(self.head.number + 1 + max_number_of_blocks, max_block_height + 1),
         )
 
         blocks = list(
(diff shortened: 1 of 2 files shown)


monitor/block_fetcher.py
```diff
@@ -195,12 +213,19 @@ def _get_block(self, blocknr_or_hash):
...
 
-    def _fetch_branch(self, max_blocks_to_fetch):
-        if max_blocks_to_fetch <= 0:
+    def _fetch_branch(self, max_blocks_to_fetch, blocknr_or_hash=None):
+
+        if blocknr_or_hash is None:
+            blocknr_or_hash = "latest"
+
+        if max_blocks_to_fetch < 0:
             raise ValueError("Maximum number of blocks to fetch must be positive")
```
jannikluhn:
```suggestion
            raise ValueError("Maximum number of blocks to fetch must not be negative")
```

monitor/block_fetcher.py
```diff
@@ -195,12 +213,19 @@ def _get_block(self, blocknr_or_hash):
 
         return block
 
-    def _fetch_branch(self, max_blocks_to_fetch):
-        if max_blocks_to_fetch <= 0:
+    def _fetch_branch(self, max_blocks_to_fetch, blocknr_or_hash=None):
```
jannikluhn: `blocknr_or_hash` isn't a very helpful name. It's not ideal either, but maybe something like `head_block_id`?

monitor/block_fetcher.py
```diff
@@ -195,12 +213,19 @@ def _get_block(self, blocknr_or_hash):
...
+
+        if blocknr_or_hash is None:
+            blocknr_or_hash = "latest"
```
jannikluhn: Should probably just be the default value of the argument.

monitor/block_fetcher.py
```diff
@@ -130,30 +136,38 @@ def fetch_and_insert_new_blocks(self, max_number_of_blocks=5000):
...
         # sync forwards at most up until the forward sync target, but no more than
         # max_number_of_blocks
-        max_forward_sync_blocks = max(
-            0,
-            min(
-                max_number_of_blocks - number_of_synced_blocks,
-                self.fetch_forward_sync_target() - self.head.number,
-            ),
+        max_forward_block_number = (
+            self.fetch_forward_sync_target()
```
jannikluhn: maybe make this a separate variable `forward_sync_target` as it's used twice in this expression

monitor/block_fetcher.py
```diff
@@ -130,30 +136,38 @@ def fetch_and_insert_new_blocks(self, max_number_of_blocks=5000):
...
         return max(self.w3.eth.blockNumber - self.max_reorg_depth, 0)
 
-    def _sync_forwards(self, max_number_of_blocks):
+    def _sync_forwards(
+        self, *, max_number_of_blocks: int, max_block_number: int
+    ) -> int:
         block_numbers_to_fetch = range(
-            self.head.number + 1, self.head.number + 1 + max_number_of_blocks
+            self.head.number + 1,
+            min(self.head.number + 1 + max_number_of_blocks, max_block_number + 1),
```
jannikluhn:
```suggestion
            min(self.head.number + max_number_of_blocks, max_block_number) + 1,
```

assuming we need the `+ 1` just because `range` is exclusive

jannikluhn:
Mostly unimportant stuff, just the number vs height thing is something we should agree on.

monitor/block_fetcher.py
```diff
@@ -117,11 +117,17 @@ def _insert_first_block(self):
         )
         self._insert_branch([block])
 
-    def fetch_and_insert_new_blocks(self, max_number_of_blocks=5000):
-        """Fetches up to `max_number_of_blocks` blocks and updates the internal state
+    def fetch_and_insert_new_blocks(
+        self, *, max_number_of_blocks=5000, max_block_number: int = None
```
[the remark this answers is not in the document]
berndbohmeier: okay

monitor/block_fetcher.py
berndbohmeier: okay

monitor/block_fetcher.py
berndbohmeier: I did not touch this, but I can change it.

monitor/block_fetcher.py
berndbohmeier: yes, I actually had this, but the problem is that the default is None for the other functions, and it made me have to write code to then change None to `latest` which felt like duplication. So if you dont mind I would leave it like this.

monitor/block_fetcher.py
berndbohmeier: okay

monitor/block_fetcher.py
berndbohmeier: I dont know, I would leave it like this, because for me it feels different for the two cases:

- The first is basically from `(head.number +1)` and number_of_blocks long
- The second is until `max_block_number` inclusively.

So for clarity I feel we should leave it like this.

monitor/block_fetcher.py
jannikluhn: alright, fine with me

monitor/block_fetcher.py
jannikluhn: ok, put like this this makes sense as well.

monitor/block_fetcher.py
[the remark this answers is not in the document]
berndbohmeier: Called it `max_block_height`, I think this is clearer.

PR #41, opened 2019-05-20
Fix/height to determine validators

berndbohmeier:


berndbohmeier:
Did not work, because someone rebased.

PR #42, opened 2019-05-21
Feature/config for e2e setup

### LINKED ISSUE #37: Create configuration file for three validator setup
chain spec, Parity configuration, enode keys, validator keys

one node will be offline, so no config for this one

there should be a precompiled validator set contract in use

weilbith:
Closes #37

I used the pre-compiled [`TestValidatorSet` contract](https://github.com/trustlines-protocol/blockchain/blob/master/contracts/contracts/TestValidatorSet.sol) for the validator definition within the chain specification file. Therefore the following list of addresses has been assigned to the `currentValidators` variable, since the `init()` function can't be called:

- `0x46ae357ba2f459cb04697837397ec90b47e48727`
- `0x260ea46e50c7111f08f7e8c863235ebcf29bb1a1`
- `0xe7a664509027ff348d379bb5d3a8340aaecb56eb`

The first two addresses relate to the signing/enode key file of `validator_one` and `validator_two`.

The directory structure is only temporally of cause and depends on the future `docker-compose` setup.


e2e_tests/configuration/chain_specification.json
```diff
@@ -0,0 +1,156 @@
...
+				"activate_at": 0,
+				"pricing": {
+					"alt_bn128_pairing": {
+						"base": 100000,
+						"pair": 80000
+					}
+				}
+			}
+		},
+		"9cce34f7ab185c7aba1b7c8140d620b4bda941d6": {
```
cducrest: What is this?

e2e_tests/configuration/chain_specification.json
```diff
@@ -0,0 +1,156 @@
+{
+	"name": "e2e_test_chain",
+	"engine": {
+		"authorityRound": {
+			"params": {
+				"stepDuration": 1,
+				"validators": {
+					"contract": "9cce34f7ab185c7aba1b7c8140d620b4bda941d7"
+				},
+				"immediateTransitions": true
```
cducrest: Why do we have immediateTransitions?

cducrest:
You are missing new lines at the end of files.

e2e_tests/configuration/chain_specification.json
weilbith: I used a "template" by Parity ([from here](https://github.com/paritytech/parity-ethereum/blob/1a16f335fac64bbfe3e6157a053d5ca6fce3d09d/ethcore/res/authority_round.json)) to not having many decisions here and only adjust/add necessary parts.
You are correct that this should be not set since our tests will not work then. :+1:

e2e_tests/configuration/chain_specification.json
weilbith: As I stated [here](https://github.com/trustlines-protocol/tlbc-monitor/pull/42#discussion_r285966047) I used a template and adjusted only necessary things. I can just remove if you don't like it. :man_shrugging:

weilbith:
@cducrest I fixed your issues. What are the concerns about new lines?

cducrest:
> @cducrest I fixed your issues. What are the concerns about new lines?

That Bernd will complain later if I don't complain now. But shouldn't all files end with a newline?

weilbith:
We don't have it currently [here](https://github.com/trustlines-protocol/blockchain/blob/master/config/trustlines-spec.json) and [here](https://github.com/trustlines-protocol/blockchain/blob/master/config/validator-config.toml). Both are the related files in the `blockchain` repository. 

For the keys I wouldn't touch anything on how they have been generated. 

A final new line character can sometimes matter in very rare cases. Sometimes they can also lead to trouble.
The formatter _Prettier_ automatically removes empty final lines (here for the `.json` type). So it would be actually more work to have them.

PR #43, opened 2019-05-22
 Add docker-compose setup for e2e tests

### LINKED ISSUE #38: Write docker compose file
The docker compose file should

- run validator nodes
- run monitor and connect to one of the validators

c.f. https://github.com/trustlines-protocol/end2end

weilbith:
Closes #38 

The configuration that the monitor depends on the validator node it connects to is enough to have no connection errors. Starting the services after each other like for the clientlib e2e tests seems not to be necessary.


e2e_tests/Dockerfile_parity
```diff
@@ -0,0 +1,3 @@
+FROM parity/parity
+
+COPY --chown=parity ./configuration/* /home/parity/.local/share/io.parity.ethereum/
```
saschagoebel: Can't this just be mounted into the default image?

e2e_tests/Dockerfile_parity
```diff
@@ -0,0 +1,3 @@
+FROM parity/parity
```
saschagoebel: As parity regularly releases betas on their official channel, I'd add a version here. 2.4.6 is the current one

e2e_tests/docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
...
+    container_name: validator_one
+    build:
+      context: .
+      dockerfile: Dockerfile_parity
+    volumes:
+      - ./validator_one/keys:/home/parity/.local/share/io.parity.ethereum/keys
+      - ./validator_one/network:/home/parity/.local/share/io.parity.ethereum/network
+    command: >
+      --engine-signer 0x46ae357ba2f459cb04697837397ec90b47e48727
+      --jsonrpc-interface all
```
saschagoebel: Either extend the Dockerfile_parity to contain this setting, or get rid of the custom image and run vanilla parity with more mounts & settings.

Personally I'd go with vanilla as it's easier to maintain.

Dockerfile
```diff
@@ -47,5 +47,6 @@ RUN apt-get update \
 FROM runner
 COPY --from=builder /opt/tlbc-monitor /opt/tlbc-monitor
 WORKDIR /opt/tlbc-monitor
+RUN mkdir reports state
```
saschagoebel: Is it possible to create those folders in the builder already? I think "RUN" adds another layer to the image.

e2e_tests/Dockerfile_parity
schmir-at-bb: If we use bind mounts, we will have to use machine images to run this on circleci (which may or may not matter to us).

saschagoebel: We're already mounting the keys into this image in the docker-compose later on, so we'll need something to run docker-compose in anyway.

weilbith: @saschagoebel I have a note for this in a commit message.
There is one downside having it as a volume which result in makes it not usable.
Docker-compose does not support `configs` like `docker stack` do. So I can only mount a whole directory. So I would have to mount this to `~/.local/share/io.parity.ethereum`, where Parity lookup for a custom configuration. Afterwards I could bind in sub-directories like I already do for the signing key etc. But Parity will use the former directory also for all its over data like its state or database. If I add this as volume to multiple containers, only one will be able to use it. The others will complain about that some files are already in use by a process.
But since all of these files are equal for all Parity nodes I would not like to copy paste them, since this would be a horror for version control and do some changes.

Any alternative good solutions?

e2e_tests/Dockerfile_parity
weilbith: Maybe better to fix this. True. :+1:

e2e_tests/docker-compose.yaml
weilbith: I could add this to the already existing Parity `config.toml` file. 
The point was more, that not each node needs to or should enable this. So I only did this for the one the monitor connects to.
Your thoughts?

Dockerfile
weilbith: Rly? That is interesting. The only ugly thing I see here is that I have to know the `WORKDIR` of the last container within the runner already. Still going for it? Possibly use an `ARG`?

e2e_tests/Dockerfile_parity
saschagoebel:
```yaml
- type: bind
        source: /path/on/host/file.txt
        target: /path/in/container/file.txt
        read_only: true
```

e2e_tests/docker-compose.yaml
saschagoebel: As said, I'd probably just move everything out of the custom parity image, so this would be fine

Dockerfile
saschagoebel: You need to know that dir in line 48 anyway ;-) If reports & state is already in the tlbc-monitor folder, you don't have to take any extra steps

e2e_tests/Dockerfile_parity
weilbith: Okay, since this is big overhead and comes with new issues of file access permissions etc. @saschagoebel and I decided to leave it as it is and open a refactor issue were to figure out if we could get rid of the custom Parity image here.

e2e_tests/docker-compose.yaml
weilbith: See [here](https://github.com/trustlines-protocol/tlbc-monitor/pull/43#discussion_r286463199)

Dockerfile
weilbith: Yeah. I don't like it as well. Doesn't matter, so I'll just shift this. :+1:

saschagoebel:
I looked at the Dockerfile_parity again and while it's possible to get around the extra image, the drawbacks are too serious to take this route right now.

See https://github.com/moby/moby/issues/2259

PR #45, opened 2019-05-22
Add e2e test modules for skip and offline reports

weilbith:
**This is WIP and has missing dependencies!!!***

First version to have some tests which evaluate the output of the monitor. This is based on _Pytest_ as the unit tests are. The tests expect the `docker-compose` setup to be running for at least 3 sealed blocks, full synced by the monitor. Setup must not run for the tests, but the monitor's `reports` directory is evaluated.

Be aware, that the monitor can not parse validator set contracts nor are all bugs be resolved. Therefore this setup is not working right now and needs some local changes to make it temporally testable.
The offline report tests have been tested with an old report file, since the offline reporter needs to be disabled due to former mentioned bugs.


e2e_tests/tests/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+@pytest.fixture(scope="session")
+def offline_validator_address():
+    return "0xe7a664509027ff348d379bb5d3a8340aaecb56eb"
+
+
+@pytest.fixture(scope="session")
+def report_directory():
+    base_directory = os.path.join(
+        os.path.dirname(os.path.realpath(__file__)), os.pardir
+    )
```
jannikluhn: Probably not necessary, but maybe cleaner to get rid of the `..` in the path by applying `os.path.realpath` once more.

e2e_tests/tests/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@pytest.fixture(scope="session")
+def skip_report_file(report_directory):
+    return os.path.join(report_directory, "skips")
+
+
+@pytest.fixture(scope="session")
+def offline_report_file(report_directory, offline_validator_address):
+    # TODO: encounter a more elegant way to handle the unknown step without need
```
jannikluhn: not sure what's wrong with the current approach, but maybe you're looking for the [glob](https://docs.python.org/3/library/glob.html) module

e2e_tests/tests/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+    # to throw here
+    for file_name in os.listdir(report_directory):
```
jannikluhn: Probably better to define the regex pattern in advance and then use a list comprehension to filter. Then you can also ensure that there's not more than one.

e2e_tests/tests/test_offline_report.py
```diff
@@ -0,0 +1,24 @@
+from os import path
+import json
+
+
+def test_offline_report_file_get_created(offline_report_file):
+    assert path.isfile(offline_report_file), "Offline report file has not been created!"
```
jannikluhn: This test can never fail, can it? Either the fixture fails or the test passes. A better approach could be to write a function `get_offline_report_files(validator)` and use it both in the fixture and in this test.

e2e_tests/tests/test_offline_report.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def test_offline_report_content_is_correct(
+    offline_report_file, offline_validator_address
+):
+    with open(offline_report_file) as file:
+        report = json.loads(file.read())
```
jannikluhn:
```suggestion
        report = json.load(file)
```

e2e_tests/tests/test_skip_report.py
```diff
@@ -0,0 +1,21 @@
+from os import path
+import re
+
+
+def test_skip_report_file_gets_created(skip_report_file):
+    assert path.isfile(skip_report_file), "Skip report file has not been created!"
```
jannikluhn: see above

e2e_tests/tests/test_offline_report.py
```diff
@@ -0,0 +1,24 @@
...
+    assert "validator" in report
+    assert report["validator"] == offline_validator_address
+    assert "missed_steps" in report
+
+    missed_steps = report["missed_steps"]
+
+    assert isinstance(missed_steps, list)
+    assert len(missed_steps) == 1
+    assert isinstance(missed_steps[0], int)
+    # TODO: compare missed step number somehow...
```
jannikluhn: probably best to compare it to the skip file

e2e_tests/tests/test_skip_report.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def test_correct_address_gets_reported_to_skip(
+    skip_report_file, offline_validator_address
+):
+    with open(skip_report_file) as file:
+        first_skip_report_entry = file.readline()
```
jannikluhn: probably better to read all lines into a list in the beginning to not have to deal with the file any longer

e2e_tests/tests/test_skip_report.py
```diff
@@ -0,0 +1,21 @@
...
+
+    assert first_skip_report_entry, "No validator has been reported to have skip!"
```
jannikluhn:
```suggestion
    assert first_skip_report_entry, "No validator has been reported to have skipped"
```

e2e_tests/tests/conftest.py
weilbith: I just shifted the existing one to the beginning. Same result. :+1:

e2e_tests/tests/conftest.py
weilbith: Yes, I think the new code with globs is better and easier to read. The missing details in the pattern (`[0-9]+_to[0-9]+`) doesn't matter I would argue.
As you somehow proposed [here](https://github.com/trustlines-protocol/tlbc-monitor/pull/45#discussion_r286790356) I return the whole list of all oflfine reports for this address. I think this is better, since I can return an empty list without need to throw here by an assert.

e2e_tests/tests/conftest.py
weilbith: I've used your suggested `glob` library for this. Is this fine?
As stated in the previous conversation I switched to return the list of all offline reports for this address.

>  Then you can also ensure that there's not more than one.

Actually I think there will be multiple reports, since for the tests, I would set the `--offline-window` to `1`. So we don't have to tackle the problem of waiting until the report happens due to a too big window. Thereby we can also check that only one step has been reported.

e2e_tests/tests/test_offline_report.py
weilbith: Yes, this is kinda useless and related to the bad fixture. Since I work with the `glob` and return a list, this should be a better case now. I think this is most-likely exactly what you propose here.

e2e_tests/tests/test_skip_report.py
weilbith: This test can fail. So I'm not sure what you refer to. Could you elaborate please? :smiley:

e2e_tests/tests/test_offline_report.py
weilbith: I see the approach. 
But the skip file is continuously, correct? So I should probably parse the first line's step... :thinking:

e2e_tests/tests/test_skip_report.py
weilbith: You mean by a fixture and then use the content for the various cases?

e2e_tests/tests/test_offline_report.py
weilbith: I've tried to implement something like this. I assume a 1-1 relation because of the offline window of size one.

e2e_tests/tests/test_skip_report.py
weilbith: I've done so and used it for the new test approaches.

weilbith:
@jannikluhn I reworked it by your feedback. Your ideas lead to some new "approaches". Let me know if you like it.

e2e_tests/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+import glob
+
+
+@pytest.fixture(scope="session")
+def offline_validator_address():
+    return "0xe7a664509027ff348d379bb5d3a8340aaecb56eb"
+
+
+@pytest.fixture(scope="session")
+def report_directory():
```
jannikluhn:
```suggestion
def report_directory_path():
```

e2e_tests/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    base_directory = os.path.realpath(
+        os.path.join(os.path.dirname(__file__), os.pardir)
+    )
+
+    return os.path.join(base_directory, "reports")
+
+
+@pytest.fixture(scope="session")
+def skip_report_file(report_directory):
```
jannikluhn:
```suggestion
def skip_report_file_path(report_directory):
```

e2e_tests/tests/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.fixture(scope="session")
+def skip_report_list(skip_report_file):
+    with open(skip_report_file) as file:
+        return file.readlines()
+
+
+@pytest.fixture(scope="session")
+def offline_report_files(report_directory, offline_validator_address):
```
jannikluhn:
```suggestion
def offline_report_file_paths(report_directory, offline_validator_address):
```

e2e_tests/tests/test_offline_report.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def get_missed_step_from_skip_report_entry(skip_report_entry):
+    missed_step_findings = re.findall(
+        "([0-9]+),0x[0-9,a-f]+,[0-9,-]+", skip_report_entry
+    )
+    missed_step_string = (
+        missed_step_findings[0] if len(missed_step_findings) == 1 else None
+    )
+    missed_step = in(missed_step_string)
```
jannikluhn:
```suggestion
    missed_step = int(missed_step_string)
```
?

jannikluhn: Also, shouldn't work with `None`

e2e_tests/tests/test_offline_report.py
```diff
@@ -0,0 +1,47 @@
...
+    assert len(offline_report_files) > 0
+
+    for file in offline_report_files:
+        assert path.isfile(file)
+
+
+def test_offline_reports_content_is_correct(
+    offline_report_files, offline_validator_address, skip_report_list
+):
+    for index, report_file in enumerate(offline_report_files):
```
jannikluhn: The test should check that there's only a single report file for each validator, so probably not loop.

e2e_tests/tests/test_skip_report.py
jannikluhn: whoops, youre right

e2e_tests/tests/test_skip_report.py
```diff
@@ -0,0 +1,19 @@
+from os import path
+import re
+
+
+def test_skip_report_file_gets_created(skip_report_file):
+    assert path.isfile(skip_report_file), "Skip report file has not been created!"
```
jannikluhn:
```suggestion
    assert path.exists(skip_report_file), "Skip report file has not been created!"
```

e2e_tests/tests/test_offline_report.py
```diff
@@ -0,0 +1,47 @@
+from os import path
+import re
+import json
+
+
+def get_missed_step_from_skip_report_entry(skip_report_entry):
+    missed_step_findings = re.findall(
+        "([0-9]+),0x[0-9,a-f]+,[0-9,-]+", skip_report_entry
```
jannikluhn: this gets used a couple of times, so maybe put it in a constant.

jannikluhn: Also, probably good to make sure that it's an exact match, without anything before or after.

weilbith:
This branch has been rebased to `master` to get the recent bugfixes as well as the ability to parse contracts. Due to that multiple errors have occur. These have been solved and can be read in detail within the related [commit message](https://github.com/trustlines-protocol/tlbc-monitor/pull/45/commits/727a774d305ca6b46cd56f01163559989921dafb).

I will continue to work on the actual tests again. At the moment there still occur some local problems with the docker-compose network. This slows me down. But I'll be up with new input soon.

weilbith:
Note: Tests fail right now because _Pytest_ has its focus on the root directory. This executes the unit tests and the end-to-end tests combined. Therefore the end-to-end test fail due to the missing running docker-compose setup.

e2e_tests/tests/test_skip_report.py
weilbith: Actually I want it more exact here. So why should I test if it is a directory or file, if it should be a file?

e2e_tests/tests/test_offline_report.py
weilbith: I would expect multiple report files for each new report, correct? I would say we don't know how long the setup is already running. It would be possible that it already run so long relative to the offline-window so that the offline validator must be reported multiple times.

I registered, that the offline-window is false. I expected the "unit" as count of missed steps. So my setup can't work. Haven't observed it due to the not working monitor until this point.

Do you agree?

e2e_tests/tests/test_offline_report.py
weilbith: > this gets used a couple of times, so maybe put it in a constant.

Because the search is different. So the pattern is the same, but the search result is slightly different.

> Also, probably good to make sure that it's an exact match, without anything before or after.

:+1:

e2e_tests/tests/test_skip_report.py
weilbith: Possibly not related to [this comment](https://github.com/trustlines-protocol/tlbc-monitor/issues/47#issuecomment-495640162) by you.

e2e_tests/tests/test_offline_report.py
jannikluhn: Validators will never be reported twice, doesn't matter for how long they are offline.

e2e_tests/tests/test_skip_report.py
jannikluhn: Then I would add this as a second condition: First does it exist, and then is it a file.

weilbith: If it helps :man_shrugging: :+1:

e2e_tests/tests/test_offline_report.py
weilbith: Yeah, I just realized it at #47 So I'll adjust the test therefore.

weilbith:
The procedure to runt the tests is the following:
- start the both validator nodes `docker-compose up validator_one validator_two`
- wait until at least `validator_one` has produced a block (else the monitor will crash)
- start the monitor service `docker-compose up monitor`
- wait until the monitor reports the offline validator
- (not necessarily here) stop all services `docker-compose down`
- run the tests scripts `pytest e2e_tests/tests`

Depending on how it works best for the _CircleCI_ setup, we can add a simple bash script for this. I think that is how it is done for the `clientlib`. I'm not 100% sure how we define the waiting for the offline report. Especially for future upcoming test scenarios. :thinking:

e2e_tests/configuration/chain_specification.json
```diff
@@ -145,11 +152,11 @@
...
...060ff1681101561190c5760b78103600184019350806020036101000a84510460018201810193505050611949565b60f860ff1681101561192957600160c060ff16820301915061194a565b60f78103600184019350806020036101000a845104600182018101935050505b5b50919050565b8280548282559060005260206000209081019282156119915760005260206000209182015b82811115611990578254825591600101919060010190611975565b5b50905061199e9190611aa2565b5090565b828054828255906000526020600020908101928215611a31579160200282015b82811115611a3057823573ffffffffffffffffffffffffffffffffffffffff168260006101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff160217905550916020019190600101906119c2565b5b509050611a3e9190611aa2565b5090565b815481835581811115611a6957818360005260206000209182019101611a689190611ae5565b5b505050565b604080519081016040528060008152602001600081525090565b604080519081016040528060008152602001600081525090565b611ae291905b80821115611ade57600081816101000a81549073ffffffffffffffffffffffffffffffffffffffff021916905550600101611aa8565b5090565b90565b611b0791905b80821115611b03576000816000905550600101611aeb565b5090565b905600a165627a7a72305820bfc92758ee832a2c427f855acd098c8060972787757af440d8e529c724f978fc0029"
```
schmir-at-bb: Has this space been added on purpose?

weilbith: No it was not intended. Seems to be ignored, since it works. ^^
I'll just remove it.

Reacting here to the docstrings tip: I would love to add them. :+1:

PR #49, opened 2019-05-24
Show proper error message if broken app state

### LINKED ISSUE #48: Ensure proper error message when state/db is broken
Ensure that we show a proper error message and hint how to fix the issue if the state can not be recovered from the db file.

berndbohmeier:
Show a proper error message
Add an experimental upgrade state mechanism to get some experience if
our approach works.

Closes #48

### DIFF
--- a/monitor/main.py
+++ b/monitor/main.py
@@ -27,7 +27,7 @@ from monitor.offline_reporter import (
     OfflineReporterStateV1,
 )
 from monitor import skip_reporter
-from monitor.skip_reporter import SkipReporter, SkipReporterStateV2, SkipReporterState
+from monitor.skip_reporter import SkipReporter, SkipReporterStateV2, SkipReporterStateV1
 from monitor.equivocation_reporter import EquivocationReporter
 from monitor.blocks import get_canonicalized_block, get_proposer, rlp_encoded_block
 from monitor.validators import (
@@ -95,7 +95,7 @@ def step_number_to_timestamp(step):
 
 class AppStateV1(NamedTuple):
     block_fetcher_state: BlockFetcherStateV1
-    skip_reporter_state: SkipReporterState
+    skip_reporter_state: SkipReporterStateV1
     offline_reporter_state: OfflineReporterStateV1
 
 
@@ -105,11 +105,11 @@ class AppStateV2(NamedTuple):
     offline_reporter_state: OfflineReporterStateV2
 
 
-def upgradeV1toV2(v1: AppStateV1):
+def upgrade_v1_to_v2(v1: AppStateV1):
     return AppStateV2(
         block_fetcher_state=v1.block_fetcher_state,
-        skip_reporter_state=skip_reporter.upgradeV1toV2(v1.skip_reporter_state),
-        offline_reporter_state=offline_reporter.upgradeV1toV2(
+        skip_reporter_state=skip_reporter.upgrade_v1_to_v2(v1.skip_reporter_state),
+        offline_reporter_state=offline_reporter.upgrade_v1_to_v2(
             v1.offline_reporter_state
         ),
     )
@@ -149,12 +149,16 @@ class App:
         self.offline_reporter = None
         self.equivocation_reporter = None
         self.initial_block_resolver = initial_block_resolver
-        self._upgrade_db = upgrade_db
 
         self._initialize_db(db_path)
         self._initialize_w3(rpc_uri)
         self._initialize_primary_oracle(chain_spec_path)
-        self._initialize_reporters(skip_rate, offline_window_size)
+
+        app_state = self._load_app_state()
+        if upgrade_db:
+            app_state = self._upgrade_app_state(app_state)
+
+        self._initialize_reporters(app_state, skip_rate, offline_window_size)
         self._register_reporter_callbacks()
         self._running = False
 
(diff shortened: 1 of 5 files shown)


monitor/main.py
```diff
@@ -86,9 +94,29 @@ def step_number_to_timestamp(step):
 
 
 class AppStateV1(NamedTuple):
-    block_fetcher_state: Any
-    skip_reporter_state: Any
-    offline_reporter_state: Any
+    block_fetcher_state: BlockFetcherStateV1
+    skip_reporter_state: SkipReporterState
```
jannikluhn:
```suggestion
    skip_reporter_state: SkipReporterStateV1
```

monitor/main.py
```diff
@@ -86,9 +94,29 @@ def step_number_to_timestamp(step):
...
+    offline_reporter_state: OfflineReporterStateV1
+
+
+class AppStateV2(NamedTuple):
+    block_fetcher_state: BlockFetcherStateV1
```
jannikluhn: Not sure about this, but maybe we should define `BlockFetcherStateV2` to keep the version numbers aligned. Otherwise this looks confusing, especially once we get a `V3`.

monitor/main.py
```diff
@@ -86,9 +94,29 @@ def step_number_to_timestamp(step):
...
+    skip_reporter_state: SkipReporterStateV2
+    offline_reporter_state: OfflineReporterStateV2
+
+
+def upgradeV1toV2(v1: AppStateV1):
```
jannikluhn:
```suggestion
def upgrade_v1_to_v2(v1: AppStateV1) -> AppStateV2:
```

Don't see a good reason why to not use snake case here.

monitor/main.py
```diff
@@ -212,9 +241,13 @@ def _initialize_primary_oracle(self, chain_spec_path: Path) -> None:
...
 
     def _initialize_reporters(self, skip_rate, offline_window_size):
-        app_state = self.db.load_pickled("appstate") or self._initialize_app_state()
-        if not isinstance(app_state, AppStateV1):
-            raise RuntimeError("app_state has unexpected format")
+        app_state = self.db.load_pickled(APP_STATE_KEY) or self._initialize_app_state()
+        if not isinstance(app_state, AppStateV2):
+            if self._upgrade_db and isinstance(app_state, AppStateV1):
+                self.logger.info("Upgrade appstate from v1 to v2")
+                app_state = upgradeV1toV2(app_state)
```
jannikluhn: To me it's surprising to have this in `_initialize_reporters`, a separate `_upgrade_db` would be clearer.

monitor/main.py
```diff
@@ -121,8 +149,9 @@ def __init__(
         self.offline_reporter = None
         self.equivocation_reporter = None
         self.initial_block_resolver = initial_block_resolver
+        self._upgrade_db = upgrade_db
```
jannikluhn: This is only used during initialization, so there's no need to store it as an instance variable.

monitor/offline_reporter.py
```diff
@@ -19,13 +19,34 @@ class OfflineInterval(NamedTuple):
...
+
+class OfflineReporterStateV2(NamedTuple):
     reported_validators: Set[bytes]
     recent_offline_intervals_by_validator: Dict[bytes, List[OfflineInterval]]
 
     # offline time in number of steps
     offline_time_by_validator: Dict[bytes, int]
 
 
+def upgradeV1toV2(v1: OfflineReporterStateV1):
```
jannikluhn:
```suggestion
def upgrade_v1_to_v2(v1: OfflineReporterStateV1):
```

monitor/main.py
berndbohmeier: That does not work, because it was named `SkipReporterState` before

berndbohmeier: I could do
```
SkipReporterState(SkipReporterStateV1):
    pass
``` 
somewhere, would you like that more?
But I need `SkipReporterState` somewhere do be backwards compatible.

monitor/main.py
berndbohmeier: I we want to keep them aligned, we should have not separated them. Keep stuff separated is for decoupling, so we should use it. 
If we do it we will at some point have something like
```
BlockFetcherStateV2(BlockFetcherStatev1):
    pass
BlockFetcherStateV3(BlockFetcherStatev2):
    pass
BlockFetcherStateV4(BlockFetcherStatev3):
    pass
....
BlockFetcherStateV10(BlockFetcherStatev9):
    pass
```
For no reason.

monitor/main.py
berndbohmeier: woops, my bad, of course this should be snake case.

monitor/main.py
berndbohmeier: Its upgrading not the entire db but only the reporter states, thats why I put it there. upgrade db is misleading too I think. 
Other suggestion?

monitor/main.py
berndbohmeier: is it a problem?

monitor/main.py
berndbohmeier: `_upgrade_reporter_states`?

monitor/main.py
jannikluhn: Argh, I see. Then I guess I'm okay with how it is right now, subclassing seems to make it worse. We *could* probably alias it with `SkipReporterStateV1 = SkipReporterState`, but this sounds like a hack too. So maybe it's best to just add a comment somewhere that this doesn't have the version suffix for historical reasons or something.

monitor/main.py
jannikluhn: nope, this example convinced me.

monitor/main.py
jannikluhn: `_upgrade_reporter_states` sounds good to me, alternatively `_upgrade_app_state` or `_migrate_db`

monitor/main.py
jannikluhn: I guess it's not super important, but the more state a class has the harder it is to understand what it actually does, to me at least (especially if it's mutable, which admittedly is not really the case here).

monitor/offline_reporter.py
berndbohmeier: done

monitor/main.py
berndbohmeier: done

monitor/main.py
berndbohmeier: aliased it

Issue #40, opened 2019-05-20
Write CircleCI config to run end-to-end scenario


jannikluhn:

weilbith:
The setup is done by #56
But this does not work stable. We need to checkout why tests fail to inconsistent report data.

PR #56, opened 2019-06-03
Add a script to run the end2end tests

schmir-at-bb:
This has been adapted from the script in the end2end repo.
@weilbith I think you should take a look at the actual tests. They sometimes do fail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -211,6 +211,24 @@ jobs:
             docker tag $LOCAL_IMAGE $DOCKER_REPO:latest
             docker push $DOCKER_REPO:latest
 
+  end2end:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - run:
+          name: Pull images
+          command: |
+            cd e2e_tests; docker-compose pull
+      - run:
+          name: Build images
+          command: |
+            cd e2e_tests; docker-compose build
+      - run:
+          name: "Run end2end tests"
+          command: |
+            e2e_tests/run-e2e.sh
+
 workflows:
   version: 2
   default:
@@ -259,3 +277,4 @@ workflows:
             - install
             - build-docker-image
           context: docker-credentials
+      - end2end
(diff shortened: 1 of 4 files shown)


e2e_tests/docker-compose.yaml
```diff
@@ -29,8 +40,24 @@ services:
...
+    container_name: testrunner
+    build:
+      context: .
+      dockerfile: Dockerfile_testrunner
+    environment:
+      - PYTHONDONTWRITEBYTECODE=1
+    volumes:
+      - tests:/work/tests
+      - reports:/work/reports
+    command: ["/opt/testrunner/bin/pytest", "tests"]
```
weilbith: Feels inconsistent. Either we use the array style like in a `Dockerfile` or we use the plain version. Please just don't start to mix them. :smile:

.circleci/config.yml
```diff
@@ -211,6 +211,24 @@ jobs:
...
 
+  end2end:
+    machine:
+      image: ubuntu-1604:201903-01
+    steps:
+      - checkout
+      - run:
+          name: Pull images
+          command: |
+            cd e2e_tests; docker-compose pull
```
weilbith: Absolutely not necessary to adjust, but I would argue this should stop after a not successful change directory (`&&`). But it will never matter.

.circleci/config.yml
```diff
@@ -211,6 +211,24 @@ jobs:
...
+      - run:
+          name: Build images
+          command: |
+            cd e2e_tests; docker-compose build
+      - run:
+          name: "Run end2end tests"
```
weilbith: Not a request, just because I'm curious and want to learn it: when does _CircleCI_ require quoted names and when not? :thinking:

weilbith:
This will close #40 and #56

weilbith:
> I think you should take a look at the actual tests. They sometimes do fail.

I can go through them again. I'll check what do not went well.

.circleci/config.yml
schmir-at-bb: You need to quote it if it's a number. Other than that, I don't know.

PR #60, opened 2019-06-04
Fix e2e setup for stable test runs

### LINKED ISSUE #40: Write CircleCI config to run end-to-end scenario

weilbith:
Closes #40 by fixing the issues of instable test runs. Checkout the commit message for information regarding the background.

Note: The time we wait in the test script is very pessimistic and could be reduced to `90` seconds. But leaving at it is should be make it absolutely sure and has no negative side effect besides the 20s delay.


weilbith:
Okay, I'm absolutely confused. I run them 13 times and all runs succeed. I need to checkout what this run on _CircleCI_ has done different.

weilbith:
2 out of 10 runs have failed running on _CircleCI_ (manually triggered to rerun the workflow). Both failed runs seem to be caused by the same issue. At least the logs look the same on a first gaze. It is hard to get further information for this scenarios. Possibly I need to increase the logging and run the workflow until it fails again.

weilbith:
Okay, I think I solved the problem. At least 40 successful _CircleCI_ runs seem to verify this.

PR #61, opened 2019-06-06
Add eip 145, 1014, 1052 in chain spec of e2e tests

cducrest:
closes https://github.com/trustlines-protocol/blockchain/issues/134

### DIFF
--- a/e2e_tests/configuration/chain_specification.json
+++ b/e2e_tests/configuration/chain_specification.json
@@ -12,7 +12,7 @@
 							]
 						},
 						"4": {
-							"contract": "9cce34f7ab185c7aba1b7c8140d620b4bda941d7"
+							"safeContract": "9cce34f7ab185c7aba1b7c8140d620b4bda941d7"
 						}
 					}
 				}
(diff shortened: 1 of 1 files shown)


e2e_tests/configuration/chain_specification.json
```diff
@@ -28,7 +28,10 @@
 		"eip140Transition": "0x0",
 		"eip211Transition": "0x0",
 		"eip214Transition": "0x0",
-		"eip658Transition": "0x0"
+		"eip658Transition": "0x0",
```
weilbith: I think this is the proposed block number for the next fork. It does not make sense to use them here at such a late block for the end-to-end tests. Or what is your intention behind this?

e2e_tests/configuration/chain_specification.json
```diff
@@ -28,7 +28,10 @@
...
+		"eip145Transition": "0x2B08BA",
```
weilbith: I think this is the proposed block number for the next fork. It does not make sense to use them here at such a late block for the end-to-end tests. Or what is your intention behind this?

cducrest: You are correct it makes no sense, I change them.

PR #62, opened 2019-06-07
Web3 five

schmir-at-bb:
This upgrades web3 to version 5


tests/fake_aura_backend.py
```diff
@@ -15,7 +15,7 @@
 from web3.datastructures import AttributeDict
 from web3.middleware import construct_formatting_middleware
 from web3.providers.eth_tester.middleware import block_key_remapper
-from web3.utils.formatters import apply_formatter_if
+from web3._utils.formatters import apply_formatter_if
```
jannikluhn: They renamed it to `_utils` to make clear that it's not public API, so we should not use it anymore. Shall we just copy the function from [here](https://github.com/ethereum/web3.py/blob/master/web3/_utils/formatters.py#L54-L59)?

schmir-at-bb: I think it does not matter. You can't rely on them not introducing breaking changes even if stuff has been documented in the official docs.

jannikluhn: I'd argue that the fact that they explicitly marked this as internal with this release shows that they intent to be more careful with breaking changed in the future.

schmir-at-bb: It looks like you care enough to fix. Can you provide a PR? (I'd be willing to take that 'risk' of keeping it as is, because we have everything pinned down)

schmir-at-bb: Beware, I may ask you to fix it in the other repositories as well :)

schmir-at-bb: It's unclear to me how to move this forward. Can we merge it as is (which would be my preference) ? Will you provide a PR? Something else?

jannikluhn: I don't know. I don't plan to submit a PR. I'm still of the opinion that we should change it, but if you disagree it's your call.

schmir-at-bb: Thanks for the feedback. I've merged it as is. We can change it later if they break it.

Issue #57, opened 2019-06-03
Script to remove a validator


cducrest:
included in the docker image of tlbc-monitor

weilbith:
Printing bytes into the equivocation log files was a horrible idea. Hex-Strings are much better and easier to handle. The problem is to have a String which is a `byte` representation by _Python_ to become `byte`s again without become parsed again. Since I don't see any advantage in logging the `byte` representation, I'll open an issue to adjust this.

weilbith:
Done by PR #65 and #66

PR #65, opened 2019-06-12
Add CLI tool to manually report validators

weilbith:
Solves  #57 except the _Docker_ integration.

Example calls:

```
report-validator report-via-file \
  --validator-set-contract-address 0x9Cc30A6088DB80F8a3B2b4d2f491AbC98559C59c \
  --equivocation-report /paht/to/tlbc-monitor/reports/equivocation_reports_for_proposer_0x505ab22ef8f3ae874dec92e60665ca490fb68192
```

---

```
report-validator report-via-arguments \
  --validator-set-contract-address 0x9Cc30A6088DB80F8a3B2b4d2f491AbC98559C59c \
  --unsigned-block-header-one 0xf901f9a08b0b6994dedb8765f7b39f95ec70a4e027812224b811cb7c95373998f3db5677a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d4934794505ab22ef8f3ae874dec92e60665ca490fb68192a02ab8e45aa54b0c26532eebe73d73b890887a047108dc9a5ce99f6cad89175b9ba056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b901000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000090ffffffffffffffffffffffffed71d01501837a120080845cc6ef929fde8302010b8f5061726974792d457468657265756d86312e33342e30826c69 \
  --signature-one 0x6f34ee0b150bfba5fc88bf9e2731318dd35776606c91cff14bf2fd2ca4a10b726d96623efdf609c11311bb852b213b233298ea3e3cce5c5c96db627bbc2ddf2900 \
  --unsigned-block-header-two 0xf901f9a0e05e23fb6a3bec793185af81000e0f723fa14205d7ac262b7b8520ed8316cbe5a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d4934794505ab22ef8f3ae874dec92e60665ca490fb68192a0c7b66672d9f26caec060c8d91b9088344eb1210efc494b2593dbdc31481ab454a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b901000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000090fffffffffffffffffffffffffffffffe05837a120080845cc6ef929fde8302010b8f5061726974792d457468657265756d86312e33342e30826c69 \
  --signature-two 0x1671ba6903ce67cca8723b40f1759afa3945bdb66e5f0b6d62bed5bfda315c4a3a68a208deea14aedbe69848d6d1448764a2db82c35d097377d3f08f93c36b7701
```


report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+    except InvalidAddressException as e:
+        raise click.BadParameter(
+            f"A address parameter is not recognized to be an address: {value}"
+        ) from e
+
+
+def validate_signature(ctx, param, value):
+    if is_hex(value) and is_0x_prefixed(value) and len(value) == 132:
+        return to_bytes(hexstr=value)
+
```
jannikluhn: feels like an unnecessary new line

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+        return validate_and_format_address(value)
+    except InvalidAddressException as e:
+        raise click.BadParameter(
+            f"A address parameter is not recognized to be an address: {value}"
+        ) from e
+
+
+def validate_signature(ctx, param, value):
+    if is_hex(value) and is_0x_prefixed(value) and len(value) == 132:
+        return to_bytes(hexstr=value)
```
jannikluhn:
```suggestion
        return decode_hex(value)
```

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+
+    else:
+        raise click.BadParameter(
+            f"A signature parameter seems to be not well structured: {value}"
+        )
+
+
+def validate_block_header(ctx, param, value):
+    if is_hex(value) and is_0x_prefixed(value):
+        return to_bytes(hexstr=value)
```
jannikluhn:
```suggestion
        return decode_hex(value)
```

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+
```
jannikluhn: ditto

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+    else:
+        raise click.BadParameter(
+            f"A block header parameter seems to be not well structured: {value}"
+        )
+
+
+def validate_equivocation_report_file(ctx, param, value):
+    try:
```
jannikluhn: I think the `try` block is way too big. Only the absolute necessary code should be in there, otherwise we risk catching wrong exceptions and the code becomes hard to follow.

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+        with open(value) as file:
+            lines = file.read().splitlines()
+
+        assert (
```
jannikluhn: IMO `assert` should only be used for tests, hacks, and checks that can never fail. If we use an exception here (see below), it should probably be `ValueError`. But here I'd just go with `if`

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+            lines[0].startswith("Proposer:")
+            and lines[1].startswith("Block step:")
+            and lines[2].startswith("Detection time:")
+        ), "Report file header does not fit!"
+
+        assert len(lines) >= 21, "Report file not long enough!"
```
jannikluhn: ditto

report-validator/report_validator/validation.py
```diff
@@ -0,0 +1,63 @@
...
+            "signature_two": validate_signature(None, None, lines[20]),
+        }
+
+    except AssertionError as e:
+        raise click.BadParameter(
+            f"The equivocation report file seems to be not well structured: {e}"
+        )
+
+    except click.BadParameter as e:
+        raise e
```
jannikluhn: don't see a reason why we should catch the exceptions if we immediately reraise it.

report-validator/setup.py
```diff
@@ -0,0 +1,15 @@
+from setuptools import setup, find_packages
+
+setup(
+    nam="report-validator",
+    setup_requires="setuptools_scm",
+    # use_scm_version=True,
```
jannikluhn: ?

report-validator/scripts/pack_contracts.py
```diff
@@ -0,0 +1,24 @@
+"""This script takes an input contract.json and filters only the needed contracts
+Usage: python pack_contracts.py input_contracts.json output_contracts.json
+"""
+import json
+
+contracts = ["ValidatorSet"]
+
+
+def pack_contracts(input_filename, output_filename):
```
jannikluhn: do we need this?

report-validator/Makefile
```diff
@@ -0,0 +1,27 @@
+VIRTUAL_ENV ?= venv
```
jannikluhn: Do we need this?

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "report-validator/blockchain"]
```
jannikluhn: do we need this?

cducrest:
I think you can have the makefile install the git submodules so that we don't freeze a version: https://github.com/trustlines-protocol/blockchain/blob/feature/bridge-deploy/bridge-deploy/Makefile#L38

report-validator/report_validator/validation.py
weilbith: Okay, which lines except the file reading can be excluded? :smiley:

weilbith:
@jannikluhn I removed the whole overhead you have questioned. By this the tools is now very slim. The whole contract part has replaced with a static and shorten ABI dictionary at the core module.

Please try to install this locally and make sure it is working. Get to the help output would be enough, so I'm confident I haven't disturbed the requirements/constraints.

weilbith:
@schmir-at-bb sry I couldn't remember this. We should possibly put such information into the task. 
:see_no_evil: 
I hope it is okay like this. The monitor as well as the report tool work both for me.
I'll squash all commits in the end.

schmir-at-bb:
`report_validator/__init__.py` is missing. 

> @schmir-at-bb sry I couldn't remember this. We should possibly put such information into the task.
> see_no_evil

no problem

weilbith:
> report_validator/__init__.py is missing.

Was just lost on my local machine.

This is not strongly related to this PR, but must the Docker image be touched due to the recent changes? The entrypoint is the `tlbc-monitor`, but theoretically `docker exec...` is enough as long as the monitor is running. Should we improve this? :thinking:

schmir-at-bb:
I think I would just upload the same image with a different entry-point. At least I don't see a good reason not to do that.

weilbith:
Well... If you think this is the best idea. :man_shrugging: 
Would you mind to grep this task? Just because this would integrate to extend our automatically uploaded images. Your are more integrated to this process.

schmir-at-bb:
What's the alternative? If we only want to have a single docker image, we should probably have a  single executable as well. In that case we should probably use click subcommands, which would be a breaking change. 

If you have a better suggestion, please speak up.

weilbith:
Good arguments. I fail to come up with a better solution. :man_shrugging:

PR #66, opened 2019-06-12
Build and upload report-validator image

schmir-at-bb:
This uploads a second image with the report-validator script as entry point to dockerhub.

### DIFF
--- a/Dockerfile.report-validator
+++ b/Dockerfile.report-validator
@@ -1,12 +1,13 @@
-# This will build the currently checked out version
+# Build a docker image for running the report-validator script
+# We just change the tlbc-monitor image to use report-validator as entrypoint.
 #
-#
-# you can build the image with:
+# Make sure the tlbc-monitor image has been build with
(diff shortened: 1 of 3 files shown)


Dockerfile.report-validator
```diff
@@ -0,0 +1,12 @@
+# This will build the currently checked out version
+#
+#
+# you can build the image with:
+#
+#    docker build -t tlbc-monitor .
```
weilbith: This lines confuses me withing that file.

weilbith:
Closes #57

Dockerfile.report-validator
schmir-at-bb: ok, I've tried to clarify the build instructions.

I'm waiting for @berndbohmeier to comment on this, since it may be controversial.

schmir-at-bb:
Bernd's not here, so I'm merging it. We can revert it later if we find a better solution.

PR #76, opened 2019-09-16
Include trustlines chain spec in docker file

jannikluhn:
At the moment, we advise to put the chain spec in a volume and mount it. This works, but only until the chain spec changes. At that point, the docker container has to be restarted manually.

By including the chain spec file in the container, the watchtower should deal with this whenever the docker image is built with a new chain spec version.

Ideally, the monitor would live in the same repository as the chain spec file so that builds are automatically triggered. But I didn't want to single-handedly without discussing it first.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ENV PIP_DISABLE_PIP_VERSION_CHECK 1
 
 RUN apt-get update \
     && apt-get install -y apt-utils python3 python3-distutils python3-dev python3-venv git \
-    build-essential \
+    build-essential curl \
     && rm -rf /var/lib/apt/lists/*
 
 RUN python3 -m venv /opt/tlbc-monitor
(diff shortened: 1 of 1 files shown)


jannikluhn:
To summarize our discussion, some people would like to avoid

- having multiple images for testnet and mainnet and
- containers being updated when the testnet chain spec file changes but mainnet is run and vice versa.

Therefore, we need a different solution than what I've done here. Options we've discussed:

- run a webserver in the parity container that serves the chain spec file to the monitor container
- use a volume that is mounted in both containers, copy the chain spec file from the parity container to the volume and then from the volume to the monitor container
- add an outer container that contains the chain spec file, launches parity and monitor within the outer container, and mount the chain spec file as a volume to both inner containers
- patch parity to allow retrieving the chain spec file or at least all historic validator set definitions via an RPC command

None of these sound very pleasant to me.

schmir-at-bb:
> use a volume that is mounted in both containers, copy the chain spec file from the parity container to the volume and then from the volume to the monitor container

it think this is just: 
- use a shared volume
- copy the configuration/chain spec file to the volume
- read the config from the volume 
- restart tlbc-monitor if the config changes

That would be my preferred solution.

PR #82, opened 2019-10-17
Pin pip and co on circleci

schmir-at-bb:
Also only setup the virtualenv after the cache has been restored.


schmir-at-bb:
> In contrast to [trustlines-protocol/contracts#262](https://github.com/trustlines-protocol/contracts/pull/262) you explicitly reference the pip in the virtual environment here. Any reason why both PRs differ here?

I think that's just because it's on PATH already in the other PR. We could change that there, because we setup the PATH, create the venv, and then make use of the fact that venv/bin is already in PATH, which is a bit surprising.

Issue #91, opened 2019-10-28
circleci: make end2end a requirement for release jobs


schmir-at-bb:
Currently running the end2end tests is not connected to anything. They can fail and we might still push out a new release.

PR #93, opened 2019-10-28
Run e2e tests on releases, they are required to publish images

cducrest:
closes https://github.com/trustlines-protocol/tlbc-monitor/issues/91


schmir-at-bb:
As said, I've done  this already.

PR #101, opened 2020-01-03
Switch to a src based layout

schmir-at-bb:
We did this on our other repos as well.

For more information on source based layout visit the blog posts
[1][2].

[1] https://blog.ionelmc.ro/2014/05/25/python-packaging/
[2] https://hynek.me/articles/testing-packaging/


schmir-at-bb:
> Why do we not switch from setup.py to setup.cfg 

yes, that is my plan. I would do that later. We also need to take care of setuptools_scm, which must be kept in the setup.py file.

